PR #2, opened 2020-03-18
Set up CI

davidor:
This PR adds everything that's needed to make the tests pass in the new test infra.


davidor:
@jmprusi The tests are passing now. I needed to change a few things.
The only important thing is that for now, I've marked our integration tests as pending. We need to figure out how to run them, but I'd rather merge this to avoid blocking other PRs.

The go coverage task is failing, but that's normal. It fails because this is the first build, so it can't compare the coverage with any previous build.

PR #8, opened 2020-04-01
Add defaulting and status initialization to the reconciler.

markusthoemmes:
These have caused confusion because of a missing webhook in the tests in this repository.


davidor:
We can delete the workaround that I added in the previous PR right?
https://github.com/knative/net-kourier/blob/master/pkg/generator/ingress_translator.go#L145

PR #10, opened 2020-04-02
Use the "go mod" style update-deps from client/tekton

mattmoor:
This uses `go mod` to pull in serving, pkg, and test-infra (with similar constraints to what are there today).

The first commit is `./hack/update-deps.sh` (lots of deletions due to `go mod tidy`)
The second commit is `./hack/update-deps.sh --upgrade` (picks up a few small things)

This is a precursor to automation that will float these forward for you daily (as you have the constraints set up for `--upgrade`)

cc @davidor


hack/tools.go
```diff
@@ -0,0 +1,8 @@
+// +build tools
+
+package tools
+
+import (
+	_ "knative.dev/test-infra"
+	_ "knative.dev/test-infra/scripts"
```
davidor, line 7: I think this is all we need from test-infra so the line above is redundant, right?

Shouldn't we also import the tests from `serving`? Something like `_ "knative.dev/serving/test/conformance/ingress"` ?

mattmoor: Yeah, `test-infra` isn't needed, I'm actually going to try to remove that `dummy.go` file upstream.

I'll play whack-a-mole with the e2e tests next.

mattmoor: Ok, I copied over the list of `required` pieces from net-contour, but to unblock the test images we need this to land: https://github.com/knative/test-infra/pull/1887. cc @chizhg

mattmoor: Build tests failed (as expected until the test-infra change), so if integration tests pass, then that should be all that's blocking this 🤞

mattmoor: there we go 💪

davidor: knative/test-infra#1887 has been merged. Can you update the PR and see if the tests pass? :pray:

mattmoor:
Pulling in the latest test-infra with `./hack/update-deps.sh --upgrade` now...

mattmoor:
Hmm, didn't work...

mattmoor:
Scratch that, I'm guessing some cache was stale.  I pushed the commit from running `--upgrade` which should pull in the fix 🤞

PR #21, opened 2020-04-03
Add PHONY targets to makefile

pmorie:
So make doesn't get confused.


davidor:
You're right @pmorie those two are PHONY. In fact, none of the targets in the Makefile are associated with a file. The rest of targets are not problematic because there aren't files with the same names as those targets.

PR #23, opened 2020-04-07
[master] Auto-update dependencies

mattmoor:
Produced via:
  `./hack/update-deps.sh --upgrade && ./hack/update-codegen.sh`
/cc davidor markusthoemmes


jmprusi:
@davidor I think we need to add a `replace` in the go.mod for the kube client, ideally for all the kube libs... give me a sec

PR #25, opened 2020-04-07
Enable HA controller setup

jmprusi:
- Leader election is enabled/disabled by adding the kourier controller name to the configmap `config-leader-election` under the key `enabledComponents`. 
- Added permissions for k8s [leases](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.17/#leaselist-v1-coordination-k8s-io) as it is used by the leader election to establish if there's an actual leader or not. 
- Changed the controller name, as it's used to create a lease with it. Leases' name can't contain uppercase characters.

### DIFF
--- a/deploy/kourier-knative.yaml
+++ b/deploy/kourier-knative.yaml
@@ -21,6 +21,47 @@ metadata:
   name: config-leader-election
   namespace: kourier-system
 data:
+  _example: |
+        ################################
+        #                              #
+        #    EXAMPLE CONFIGURATION     #
+        #                              #
+        ################################
+
+        # This block is not actually functional configuration,
+        # but serves to illustrate the available configuration
+        # options and document them in a way that is accessible
+        # to users that `kubectl edit` this config map.
+        #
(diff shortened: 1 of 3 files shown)


deploy/kourier-knative.yaml
```diff
@@ -16,6 +16,14 @@ metadata:
   namespace: kourier-system
 ---
 apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: config-leader-election
+  namespace: kourier-system
+data:
+  enabledComponents: kourier
```
pmorie, line 24: Personally, it's fine with me to enable this by default; in general the convention for configmaps like this seems to be to include an explanation of what's in the file. Example from previous PR: https://github.com/knative/serving/pull/6683/files#diff-75b02e0c469b259e0eea8dd0e40c3dd7

jmprusi: @pmorie please, review my latest commit. Thanks for you help enabling controller HA to kourier.

PR #31, opened 2020-04-12
Disable TestRetry to unblock releases.

mattmoor:
We have our release coming up Tuesday, and if we want to produce one for kourier, then we need to disable this test.

### DIFF
--- a/test/e2e-tests.sh
+++ b/test/e2e-tests.sh
@@ -21,6 +21,8 @@ initialize $@  --skip-istio-addon
 
 go_test_e2e -timeout=20m -parallel=12 \
 	    ./vendor/knative.dev/serving/test/conformance/ingress \
+            `# TODO(#30): TestRetry is blocking the nightly release.` \
+	     -run="Test[^R]" \
 	    --ingressClass=kourier.ingress.networking.knative.dev || fail_test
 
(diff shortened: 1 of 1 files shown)


test/e2e-tests.sh
```diff
@@ -21,6 +21,8 @@ initialize $@  --skip-istio-addon
 
 go_test_e2e -timeout=20m -parallel=12 \
 	    ./vendor/knative.dev/serving/test/conformance/ingress \
+            `# TODO(#30): TestRetry is blocking the nightly release.` \
+	     -run="Test[^R]" \
```
davidor, line 25: I think it would be better to make this more explicit to avoid excluding other tests by accident.
Shouldn't we just use `TestRetry` instead here?

mattmoor: Ack this is crude, but Go only provides a way to specify which tests to run (not which *not* to run), and this is the best I've come up with for within those constraints.  Definitely open to suggestions, as we have similar crudeness in net-contour. :(

davidor: Ah I see, it doesn't support negative lookaheads.
I guess that's good enough until we find why this test fails on nightly :+1:

PR #34, opened 2020-04-13
Use ko in deployment yaml

davidor:


davidor:
@mattmoor  I think this should fix the error we're seeing in the nightly build.

davidor:
You are right @mattmoor

PR #39, opened 2020-04-15
Validate Ingress before adding it to the cache: Avoid duplicated domains

jmprusi:


pkg/generator/generator.go
```diff
@@ -82,7 +82,9 @@ func addIngressToCaches(caches *Caches,
 	if err != nil {
 		return err
 	}
-	if ingressTranslation != nil {
+
+	// If the translated ingress is not nil, and gets validated succesfully we add it to the config cache.
```
mattmoor, line 86: Found potential misspelling(s):
```suggestion
	// If the translated ingress is not nil, and gets validated successfully we add it to the config cache.
```

mattmoor:
/hold
needs to be on its own line 😬

mattmoor:
oh! maybe not anymore?!?!

jmprusi:
> oh! maybe not anymore?!?!

I wasn't sure! but the bot actually added the `do-not-merge/hold` :)

pkg/reconciler/ingress/ingress.go
```diff
@@ -67,8 +70,16 @@ func (reconciler *Reconciler) Reconcile(ctx context.Context, key string) error {
 	ingress := original.DeepCopy()
 	ingress.SetDefaults(ctx)
 	ingress.Status.InitializeConditions()
-
-	if err := reconciler.updateIngress(ingress); err != nil {
+	// This is the generation we are going to handle during this reconciliation loop.
+	ingress.Status.ObservedGeneration = ingress.GetGeneration()
+
+	err = reconciler.updateIngress(ingress)
+	if err != nil && err == errors.ErrDomainConflict {
```
markusthoemmes, line 77: You can drop the `!= nil` check here.

pkg/generator/generator.go
```diff
@@ -82,8 +82,14 @@ func addIngressToCaches(caches *Caches,
 	if err != nil {
 		return err
 	}
+
+	// If the translated ingress is not nil, and gets validated successfully we add it to the config cache.
 	if ingressTranslation != nil {
-		caches.AddTranslatedIngress(ingress, ingressTranslation)
+		if err := caches.ValidateIngress(ingressTranslation); err == nil {
+			caches.AddTranslatedIngress(ingress, ingressTranslation)
+		} else {
```
markusthoemmes, line 90: Let's turn this around to make it a bit more idiomatic:

```golang
if ingressTranslation == nil {
  return nil
}
if err := caches.ValidateIngress(ingressTranslation); err != nil {
  return err
}

caches.AddTranslatedIngress(ingress, ingressTranslation)
return nil
```

pkg/errors/errors.go
```diff
@@ -0,0 +1,7 @@
+package errors
```
markusthoemmes, line 1: Let's not make this a new package. I think it's just fine defining this in the `caches.go` file and use it via `generator.ErrDomainConflict` in the reconciler. I'm personally not a fan of centralized errors like this and would rather keep them close to their implementation.

pkg/config/config.go
```diff
@@ -34,6 +34,9 @@ const (
 	ExtAuthzTimeout          = "KOURIER_EXTAUTHZ_TIMEOUT"
 	ExternalAuthzCluster     = "extAuthz"
 
+	DuplicatedDomainReason  = "DomainConflict"
+	DuplicatedDomainMessage = "Ingress rejected as its domain conflicts with another ingress"
```
markusthoemmes, line 38: This doesn't feel like config 🤔 , since these values are unused, shall we just inline the strings for now?

jmprusi:

TestUpdate failed. Retrying to get more info.

pkg/generator/generator.go
```diff
@@ -82,10 +82,14 @@ func addIngressToCaches(caches *Caches,
...
 		return err
 	}
-	if ingressTranslation != nil {
-		caches.AddTranslatedIngress(ingress, ingressTranslation)
-	}
 
+	if ingressTranslation == nil {
+		return nil
+	}
+	if err := caches.ValidateIngress(ingressTranslation); err != nil {
```
davidor, line 89: I think it would be better to make `ValidateIngress` private and call it first thing in the `AddTranslatedIngress` function.
The reason is that as it is now, the user of the `caches` needs to remember to always call `ValidateIngress` before calling `AddTranslatedIngress`. This might cause bugs later if we need to call `AddTranslatedIngress` from a different part of the codebase.

pkg/generator/caches_test.go
```diff
@@ -185,6 +185,40 @@ func createTestDataForIngress(caches *Caches,
...
+		ingressNamespace:     "ingress_2_namespace",
+		routes:               []*route.Route{{Name: "route_for_ingress_2"}},
+		clusters:             []*v2.Cluster{{Name: "cluster_for_ingress_2"}},
+		externalVirtualHosts: []*route.VirtualHost{{Name: "external_host_for_ingress_2", Domains: []string{"external_host_for_ingress_2"}}},
+		//This domain should clash with the cached ingress.
+		internalVirtualHosts: []*route.VirtualHost{{Name: "internal_host_for_ingress_2", Domains: []string{"internal_host_for_ingress_1"}}},
+	}
+
+	err := caches.ValidateIngress(&translatedIngress)
+	if err == nil {
```
davidor, line 216: I think we should check that this returns `ErrDomainConflict` otherwise we could return any other error and the test would pass anyway.

pkg/reconciler/ingress/ingress.go
```diff
@@ -67,8 +67,16 @@ func (reconciler *Reconciler) Reconcile(ctx context.Context, key string) error {
...
-	if err := reconciler.updateIngress(ingress); err != nil {
+	// This is the generation we are going to handle during this reconciliation loop.
+	ingress.Status.ObservedGeneration = ingress.GetGeneration()
+
+	err = reconciler.updateIngress(ingress)
+	if err == generator.ErrDomainConflict {
+		// If we had an error due to a duplicated domain, we must mark the ingress as failed with a
+		// custom status. We don't want to return an error in this case as we want to update its status.
+		reconciler.logger.Errorw(err.Error(), ingress.Name, ingress.Namespace)
+		ingress.Status.MarkLoadBalancerFailed("DomainConflict", "Ingress rejected as its domain conflicts with another ingress")
```
davidor, line 78: Will this be retried at some point or the ingress stays there forever marked as invalid? :thinking:

pkg/generator/caches_test.go
jmprusi: There's no other error that could happen inside ValidateIngress /now/. But I will add that check in case we extend that validation function.

pkg/reconciler/ingress/ingress.go
markusthoemmes: A global resync would retry it, I guess?

jmprusi: Yes, the global resync will retry it. Do we want that to happen?

markusthoemmes: Yep, I think that's fine. If the conflict is resolved this may pass happily.

pkg/generator/generator.go
```diff
@@ -86,10 +86,10 @@ func addIngressToCaches(caches *Caches,
 	if ingressTranslation == nil {
 		return nil
 	}
-	if err := caches.ValidateIngress(ingressTranslation); err != nil {
+
+	if err := caches.AddTranslatedIngress(ingress, ingressTranslation); err != nil {
```
markusthoemmes, line 90: nit: `return caches.AddTranslatedIngress(ingress, ingressTranslation)`

pkg/generator/caches_test.go
```diff
@@ -212,8 +212,8 @@ func TestValidateIngress(t *testing.T) {
 		internalVirtualHosts: []*route.VirtualHost{{Name: "internal_host_for_ingress_2", Domains: []string{"internal_host_for_ingress_1"}}},
 	}
 
-	err := caches.ValidateIngress(&translatedIngress)
-	if err == nil {
+	err := caches.validateIngress(&translatedIngress)
+	if err == ErrDomainConflict {
```
davidor, line 216: This should be `!=` right? We could also use `assert.Error(t, err, ErrDomainConflict)`

PR #42, opened 2020-04-17
Remove unnecesary XDS snapshot update

jmprusi:
The OnEviction callback will take care of updating the XDS when the cluster is already removed.

### DIFF
--- a/pkg/reconciler/ingress/ingress.go
+++ b/pkg/reconciler/ingress/ingress.go
@@ -93,17 +93,7 @@ func (reconciler *Reconciler) deleteIngress(namespace, name string) error {
 		reconciler.statusManager.CancelIngressProbing(ingress)
 	}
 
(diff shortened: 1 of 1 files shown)


davidor:
The OnEviction callback runs after a few seconds. It's set to 15 now. This means that all that information related with the ingress just removed will not be deleted from the Envoy config until at least 15 seconds later. I guess that could cause some issues, or am I missing something? :thinking:

jmprusi:
> The OnEviction callback runs after a few seconds. It's set to 15 now. This means that all that information related with the ingress just removed will not be deleted from the Envoy config until at least 15 seconds later. I guess that could cause some issues, or am I missing something? 🤔

There will be a dangling route pointing to the cluster. All of this, if there's no other event, will get removed after the 15seconds (OnEviction). If there's a request, the Activator will return 404 as the ingress doesn't exist in Knative anymore. So, yes, we will route an unexisting knative ingress for 15s, but that's not something that really bothers me, as it will get ""garbage collected"" after 15 secs :)

PR #44, opened 2020-04-17
Remove unused http gateway

jmprusi:
Doing some cleaning, we left this at some moment and never removed it...


jmprusi:

testUpdate failed with a 503...

PR #45, opened 2020-04-18
Reduce unknown cluster retries

jmprusi:


jmprusi:
After checking the integration test output, I didn't found any: `code: 13 message route: unknown weighted cluster"`

let's give it another try: 

pkg/envoy/xds_server.go
```diff
@@ -96,12 +96,11 @@ func (envoyXdsServer *XdsServer) RunManagementServer() {
...
 }
 
-func (envoyXdsServer *XdsServer) SetSnapshot(snapshot *cache.Snapshot, nodeID string) error {
-	err := envoyXdsServer.snapshotCache.SetSnapshot(nodeID, *snapshot)
-
-	if err != nil {
-		return err
-	}
+func (envoyXdsServer *XdsServer) GetSnapshot() (cache.Snapshot, error) {
+	//TODO: node id as a const or func arg?
```
davidor, line 100: I think it's better as a function argument like it is now in the `SetSnapshot` function. That will allow us to keep the `envoy` package without kourier-only dependencies.

pkg/reconciler/ingress/ingress.go
```diff
@@ -83,6 +83,29 @@ func (reconciler *Reconciler) Reconcile(ctx context.Context, key string) error {
...
+	currentSnapshot, err := reconciler.EnvoyXDSServer.GetSnapshot()
+	if err != nil {
+		return err
+	}
+	newSnapshot, err := reconciler.CurrentCaches.ToEnvoySnapshot()
+	if err != nil {
+		return err
+	}
+
+	// Let's update the list of clusters with the previous and the new clusters and
```
davidor, line 97: I don't understand why this step is needed. We were sending a randomly generated header before to make sure something changed. Does that mean that this would work with anything, or is there something special about the list of clusters?

jmprusi: The randomly generated header is a "hack".

What we are doing here is to:

* Update the snapshot with the cluster list, with the resource version bumped, but keeping the listener/routes with the previous version. So envoy just updates the CDS part first and starts the warming them. 

* Then, send the NEW snapshot, that contains the new listeners/routes... with also, the "previous" cluster list. All with the same resource version. Making sure the clusters are already there, and thus avoiding the RDS service to fail with "unknown clusters".

That means not having to retry the config generation. 

I want to run some more tests: Letting this run for some hours in a cluster to ensure the IngressNotConfigured issue doesn't manifest.

davidor: I see. So this tries to "warm" a cluster before anyone points to it. I wonder why Envoy doesn't handle that internally but if it works, :+1:
I think we should add a comment here explaining this thing about warmed clusters and avoiding retries.

jmprusi: yup! :) I will run some tests, put some more comments, and remove the wip.

PR #47, opened 2020-04-21
Filter global resyncs to only use kourier class too.

markusthoemmes:
Currently, our global resyncs don't respect the kourier class, leading to weird behavior if multiple ingress providers are being used.

/assign @jmprusi @davidor


jmprusi:

503 in testupdate

PR #48, opened 2020-04-21
Make use of generated reconciler architecture.

markusthoemmes:
This became a little bigger than I wanted initially, let me know if you want me to break this down into smaller chunks please.

This makes use of the "central" generated reconciler architecture to safe us from quite a bit of boilerplate and make the usage of certain concepts easier.

The only semantic change in here is the use of a finalizer to remove the ingress from our config. Arguably that is preferable over reacting to "NotFound" as I don't think we're technically guaranteed to even see that event.


jmprusi:
@markusthoemmes from the logs of the failed integration tests, I can see: 

```User \"system:serviceaccount:kourier-system:3scale-kourier\" cannot patch resource \"ingresses\" in API group \"networking.internal.knative.dev\" in the namespace \"serving-tests\""}```

jmprusi:
/approve

As this is a big change, let's wait for @davidor to lgtm it also.

markusthoemmes:

TestUpdate seems to fail fairly often?

jmprusi:
> test pull-knative-net-kourier-integration-tests
> 
> TestUpdate seems to fail fairly often?

Yes, it is failing more than the usual lately... we will need to review it...

PR #57, opened 2020-05-04
Multiple controller threads support

jmprusi:
- Remove DefaultThreadsPerController override 
- Expose only the required cache functions.
- Add locks to exported cache functions to allow multiple controllers workers
- New func cache.InitConfig to initialize the configuration 
- Fix unit tests by updating go-cmp package.

### DIFF
--- a/pkg/generator/caches.go
+++ b/pkg/generator/caches.go
@@ -48,14 +48,16 @@ type Caches struct {
 	logger              *zap.SugaredLogger
 }
 
-func NewCaches(logger *zap.SugaredLogger) *Caches {
-	return &Caches{
+func NewCaches(logger *zap.SugaredLogger, kubernetesClient kubeclient.Interface, extAuthz bool) (*Caches, error) {
+	c := &Caches{
 		ingresses:           make(map[string]*v1alpha1.Ingress),
 		translatedIngresses: make(map[string]*translatedIngress),
 		clusters:            newClustersCache(logger.Named("cluster-cache")),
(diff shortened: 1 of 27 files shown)


pkg/reconciler/ingress/controller.go
```diff
@@ -148,8 +148,7 @@ func NewController(ctx context.Context, cmw configmap.Watcher) *controller.Impl
 	// until a Knative service is deployed. This is important because the
 	// gateway pods will not be marked as healthy until they have been able to
 	// fetch a config.
-	r.caches.AddStatusVirtualHost()
-	err := r.caches.SetListeners(kubernetesClient)
+	err := r.caches.InitConfig(kubernetesClient, r.extAuthz)
```
davidor, line 151: Should the initialization be part of `caches.new()` ?
Otherwise, it forces the caller to call InitConfig always after caches.new(), something that we might forget at some point.

pkg/generator/caches.go
```diff
@@ -56,13 +58,38 @@ func NewCaches(logger *zap.SugaredLogger) *Caches {
 	}
 }
 
+func (caches *Caches) UpdateIngress(ingress *v1alpha1.Ingress, ingressTranslation *translatedIngress, kubeclient kubeclient.Interface) error {
+	// we hold a lock for Updating the ingress, to avoid another worker to generate an snapshot just when we have
+	// deleted the ingress before adding it.
+	caches.mu.Lock()
```
davidor, line 64: I think that `GetIngress()` below should be protected as well.

PR #58, opened 2020-05-05
Wait until the static IP is hooked up to the Envoy pod

tcnghia:
There seem to be a recent change in GKE that after the static IP is available we may still get many connection refused/connection timeout errors, resulting in very flaky TestRetry (https://testgrid.knative.dev/net-kourier#Summary&width=5) and failed nightly builds.

This PR adds some more waiting in the test setup until we see that we can make HTTP connection to the static IP.  It seems to resolve the same flakiness in knative.dev/net-istio.


jmprusi:
Thanks @tcnghia ! I think that updating the go-control-plane package breaks kourier right now, can you try using the 0.9.1 version of that package ?

tcnghia:
@jmprusi that works.

I had to add a bunch of other pins as well.  I think those are the reasons the auto-dep update isn't working well.

jmprusi:
Thanks @tcnghia !

/approve
/lgtm

PR #59, opened 2020-05-05
Remove logrus dependency and use common logger.

markusthoemmes:
This dependency seemed arbitrary and completely unneeded. So why not remove some code? :)


jmprusi:
@markusthoemmes thanks! that's some leftover code we had there.. also the unit tests fails due to go-cmp (if you use the 4.0 this is fixed).

Let me try to get a PR merged that fixes this, and then you can rebase.

markusthoemmes:
Rebased!

PR #64, opened 2020-05-12
Remove 30s global resync

jmprusi:
This PR removes a resync that runs every 30s that is not needed anymore as it can cause problems on big clusters with thousands of ksvcs.


jmprusi:
/hold 

running tests and making sure it works as expected.

jmprusi:
/unhold

tested, and working properly

PR #66, opened 2020-05-13
Fix module usage in presubmit tests and update scripts.

markusthoemmes:
The presubmit tests don't seem to use the vendor directory. This fixes that.


davidor:
The PR looks good @markusthoemmes 
I think the build is failing simply because the helloworld image is not imported in https://github.com/knative/net-kourier/blob/master/hack/tools.go

PR #70, opened 2020-05-16
Change go.mod to use 1.13, so ingress tests actually runs

tcnghia:
Currently there is a bug in test-infra such that when no integration tests is run we won't see a failure.

Using go 1.14 in go.mod hits this condition. See https://prow.knative.dev/view/gcs/knative-prow/pr-logs/pull/knative_net-kourier/69/pull-knative-net-kourier-integration-tests/1261318375200526337

Change back to 1.13 so tests actually will run.


tcnghia:
/close
changing back to 1.13 doesn't seem to help

PR #72, opened 2020-05-18
Warm the internal config cache to avoid sending an empty snapshot.

jmprusi:
Add a WaitForSync method in caches that exposes a channel that will receive an event once the cache has been warmed. 

* The cache pkg can get initialized with a list of Ingresses that will be used to determine if the cache has been warmed (all the ingresses in that list should be in the cache).

* The xdsServer waits for the cache to be warmed before starting to listen. This is also controlled by a timeout of 12 minutes (~1000 ksvcs)


pkg/readiness/ready_server.go
```diff
@@ -0,0 +1,94 @@
...
+	"knative.dev/net-kourier/pkg/config"
+	"knative.dev/net-kourier/pkg/generator"
+	"knative.dev/serving/pkg/apis/networking"
+	v1alpha12 "knative.dev/serving/pkg/client/clientset/versioned/typed/networking/v1alpha1"
+)
+
+const (
+	// The port used by the readiness endpoint.
+	readinessEndpointPort = 9191
+	// readinessEndpointMaxIter the max number of attemps to match the ingress count.
```
mattmoor, line 20: Found potential misspelling(s):
```suggestion
	// readinessEndpointMaxIter the max number of attempts to match the ingress count.
```

jmprusi:
@davidor @markusthoemmes can you give a look at this PR? thanks! :)

pkg/generator/caches.go
```diff
@@ -45,20 +45,26 @@ type Caches struct {
...
 		clustersToIngress:   make(map[string][]string),
 		logger:              logger,
+		ingressesToSync:     ingressesToSync,
 	}
 	err := c.initConfig(kubernetesClient, extAuthz)
 	return c, err
 }
 
+func (caches *Caches) HasSynced() bool {
+	return len(caches.ingressesToSync) == 0
```
markusthoemmes, line 65: does this not need to be behind a lock?

pkg/reconciler/ingress/controller.go
```diff
@@ -168,3 +189,35 @@ func NewController(ctx context.Context, cmw configmap.Watcher) *controller.Impl
 
 	return impl
 }
+
+func waitForCache(caches *generator.Caches) {
```
markusthoemmes, line 193: Why not drive this by events? Have we considered adding an "observer" pattern to the cache that informs us every time there's a state change to `ingressesToSync`?

pkg/reconciler/ingress/controller.go
```diff
@@ -168,3 +189,35 @@ func NewController(ctx context.Context, cmw configmap.Watcher) *controller.Impl
...
+
+func getReadyIngresses(knativeClient v1alpha12.NetworkingV1alpha1Interface) (map[string]struct{}, error) {
+	ingresses, err := knativeClient.Ingresses("").List(metav1.ListOptions{})
+	if err != nil {
+		return nil, err
+	}
+	ingressesToWarm := map[string]struct{}{}
+	for _, ingress := range ingresses.Items {
+		if val, ok := ingress.Annotations[networking.IngressClassAnnotationKey]; ok {
+			if val == config.KourierIngressClassName && ingress.GetStatus().GetCondition(v1alpha1.IngressConditionNetworkConfigured).IsTrue() {
```
markusthoemmes, line 217:
```suggestion
		if ingress.Annotations[networking.IngressClassAnnotationKey] == config.KourierIngressClassName && ingress.GetStatus().GetCondition(v1alpha1.IngressConditionNetworkConfigured).IsTrue() {
```

Has the same semantics I think.

pkg/reconciler/ingress/controller.go
```diff
@@ -168,3 +189,35 @@ func NewController(ctx context.Context, cmw configmap.Watcher) *controller.Impl
...
+	timeout := time.After(cacheWarmUPTimeout)
+	tick := time.Tick(1 * time.Second)
+pollCacheInSync:
+	for {
+		select {
+		case <-timeout:
+			break pollCacheInSync
+		case <-tick:
+			if caches.HasSynced() {
+				break pollCacheInSync
```
markusthoemmes, line 203: you can just `return` here, I think.

pkg/reconciler/ingress/controller.go
```diff
@@ -168,3 +189,35 @@ func NewController(ctx context.Context, cmw configmap.Watcher) *controller.Impl
 
 	return impl
 }
+
+func waitForCache(caches *generator.Caches) {
+	timeout := time.After(cacheWarmUPTimeout)
+	tick := time.Tick(1 * time.Second)
```
markusthoemmes, line 195: This leaks a timer.

```suggestion
	tick := time.NewTicker(1 * time.Second)
	defer tick.Stop()
```

pkg/reconciler/ingress/controller.go
```diff
@@ -49,22 +54,30 @@ const (
 	gatewayLabelKey   = "app"
 	gatewayLabelValue = "3scale-kourier-gateway"
 
-	nodeID         = "3scale-kourier-gateway"
-	gatewayPort    = 19001
-	managementPort = 18000
+	nodeID             = "3scale-kourier-gateway"
+	gatewayPort        = 19001
+	managementPort     = 18000
+	cacheWarmUPTimeout = 720 * time.Second
```
davidor, line 60: This seems to be a long timeout. How many ingresses should kourier need to reconcile to take more than this? :thinking:

jmprusi: The tests show that resyncing  ~1000 ksvcs could take a long time.

pkg/generator/caches.go
```diff
@@ -45,20 +45,28 @@ type Caches struct {
 	listeners           []*v2.Listener
 	statusVirtualHost   *route.VirtualHost
 	logger              *zap.SugaredLogger
+	ingressesToSync     map[string]struct{}
 }
 
-func NewCaches(logger *zap.SugaredLogger, kubernetesClient kubeclient.Interface, extAuthz bool) (*Caches, error) {
+func NewCaches(logger *zap.SugaredLogger, kubernetesClient kubeclient.Interface, extAuthz bool, ingressesToSync map[string]struct{}) (*Caches, error) {
```
davidor, line 51: This moves some complexity to the callers. Now they need to know how to build the keys that represent an ingress, which should be internal to this module. Shouldn't this method receive references to ingresses instead?

pkg/reconciler/ingress/controller.go
```diff
@@ -168,3 +189,33 @@ func NewController(ctx context.Context, cmw configmap.Watcher) *controller.Impl
...
 }
+
+func waitForCache(caches *generator.Caches) {
+	timeout := time.After(cacheWarmUPTimeout)
+	tick := time.NewTicker(1 * time.Second)
+	defer tick.Stop()
+	for {
+		select {
+		case <-timeout:
+			return
```
davidor, line 200: I think it would be good to log a warning or an error here.

pkg/reconciler/ingress/controller.go
```diff
@@ -168,3 +191,34 @@ func NewController(ctx context.Context, cmw configmap.Watcher) *controller.Impl
...
+}
+
+func getReadyIngresses(knativeClient v1alpha12.NetworkingV1alpha1Interface) (map[string]struct{}, error) {
+	ingresses, err := knativeClient.Ingresses("").List(metav1.ListOptions{})
+	if err != nil {
+		return nil, err
+	}
+	ingressesToWarm := map[string]struct{}{}
+	for _, ingress := range ingresses.Items {
+		if ingress.Annotations[networking.IngressClassAnnotationKey] == config.KourierIngressClassName && ingress.GetStatus().GetCondition(v1alpha1.IngressConditionNetworkConfigured).IsTrue() {
```
davidor, line 219: Not too important, but I think we could simplify this a bit extracting 2 helpers to `ingress.go`.
We could have something like: `IsManagedByKourier(ingress)` and `HasNetworkingConfigured(ingress)`

jmprusi: Yes, let's add that in another PR.

pkg/generator/caches.go
jmprusi: true, check the latest changes.

pkg/reconciler/ingress/controller.go
davidor: :+1:

davidor:
Looks good :+1: but maybe we should merge #89 first. The integration tests did not run in this PR.

jmprusi:
@davidor thanks! But I will make it more based on events instead of polling as per @markusthoemmes suggestion! 

/hold

pkg/generator/caches.go
```diff
@@ -45,20 +45,54 @@ type Caches struct {
...
+		synced:              make(chan bool),
 	}
 	err := c.initConfig(kubernetesClient, extAuthz)
+
+	// Listen for Sync events, triggered when there's a removal of an object of the "IngressesToSync" array
+	go c.syncEvents()
+
+	if ingressesToSync == nil {
+		// If ingressesToSync is empty, we send an event to the "synced" channel as we don't need to warm anything.
+		c.synced <- true
```
markusthoemmes, line 68: Just close the channel here. The semantics are exactly the same as sending an event + closing it afterwards.

pkg/generator/caches.go
```diff
@@ -45,20 +45,54 @@ type Caches struct {
...
 		translatedIngresses: make(map[string]*translatedIngress),
 		clusters:            newClustersCache(logger.Named("cluster-cache")),
 		clustersToIngress:   make(map[string][]string),
 		logger:              logger,
+		synced:              make(chan bool),
 	}
 	err := c.initConfig(kubernetesClient, extAuthz)
+
+	// Listen for Sync events, triggered when there's a removal of an object of the "IngressesToSync" array
+	go c.syncEvents()
```
markusthoemmes, line 64: With the comment below in mind, this is not necessary I think.

pkg/generator/caches.go
```diff
@@ -45,20 +45,54 @@ type Caches struct {
...
 
-func NewCaches(logger *zap.SugaredLogger, kubernetesClient kubeclient.Interface, extAuthz bool) (*Caches, error) {
+func NewCaches(logger *zap.SugaredLogger, kubernetesClient kubeclient.Interface, extAuthz bool, ingressesToSync []*v1alpha1.Ingress) (*Caches, error) {
 	c := &Caches{
 		ingresses:           make(map[string]*v1alpha1.Ingress),
 		translatedIngresses: make(map[string]*translatedIngress),
 		clusters:            newClustersCache(logger.Named("cluster-cache")),
 		clustersToIngress:   make(map[string][]string),
 		logger:              logger,
+		synced:              make(chan bool),
```
markusthoemmes, line 59: If you're just closing this, make this `chan struct{}`

pkg/generator/caches.go
```diff
@@ -204,6 +241,18 @@ func (caches *Caches) ToEnvoySnapshot() (cache.Snapshot, error) {
...
 }
 
+func (caches *Caches) deleteFromSyncList(ingressName, ingressNamespace string) {
+	// If caches are not synced, we try to delete the ingress from the IngressesToSync list
+	if !caches.hasSynced() {
+		delete(caches.ingressesToSync, mapKey(ingressName, ingressNamespace))
+
+		// Now let's see if after the delete we are in Sync and send the event.
+		if caches.hasSynced() {
+			caches.synced <- true
```
markusthoemmes, line 251: Just close here too.

pkg/reconciler/ingress/controller.go
```diff
@@ -168,3 +191,32 @@ func NewController(ctx context.Context, cmw configmap.Watcher) *controller.Impl
...
+		}
+	}
+}
+
+func getReadyIngresses(knativeClient v1alpha12.NetworkingV1alpha1Interface) ([]*v1alpha1.Ingress, error) {
+	ingresses, err := knativeClient.Ingresses("").List(metav1.ListOptions{})
+	if err != nil {
+		return nil, err
+	}
+	var ingressesToWarm []*v1alpha1.Ingress
```
markusthoemmes, line 214: I'd preallocate this to the size of all ingresses, like:

```
ingressesToWarm := make([]*v1alpha1.Ingress, 0, len(ingresses.Items))
```

pkg/reconciler/ingress/controller.go
```diff
@@ -129,15 +143,24 @@ func NewController(ctx context.Context, cmw configmap.Watcher) *controller.Impl
...
-	}
+	// Let's start the management server when our cache is in sync, to avoid sending an incomplete configuration
+	// to an already running gateway container. If the cache is not warmed up after "cacheWarmUPTimeout" we just
+	// start the server as somehow we couldn't sync.
+	go func() {
+		waitForCache(logger, caches)
+
+		snapshot, err := r.caches.ToEnvoySnapshot()
+		if err != nil {
+			panic(err)
```
markusthoemmes, line 154: Use `logger.Fatalw("Failed to create snapshot", zap.Error(err))`. Same suggestion below.

pkg/generator/caches.go
jmprusi: I didn't know about that behavior! thanks, that's really useful! :)

jmprusi:
> A test for the cache behavior would be great.

added some basic tests that should cover the cache warming.

jmprusi:
@markusthoemmes @davidor can I get a review? thanks!

PR #75, opened 2020-05-19
Aligning test call with net-istio.

markusthoemmes:
Our integration tests are still broken. Trying to flesh out what exactly is causing it.

/hold

We know it worked if the integration test suite actually runs. See https://prow.knative.dev/view/gcs/knative-prow/pr-logs/pull/knative_net-kourier/74/pull-knative-net-kourier-integration-tests/1262687126735032320 for the failure.


markusthoemmes:
/hold cancel

This might be fair enough for now. It fixes the tests!

PR #82, opened 2020-05-22
[WIP] Replace loadbalancer service with a ClusterIP service

jmprusi:


jmprusi:
just checking something.

PR #89, opened 2020-06-01
Re-enable ingress conformance

dprotaso:


dprotaso:
@davidor this should be good now - let me know if you have questions about the changes

PR #96, opened 2020-06-09
Preallocate slices.

markusthoemmes:
Most of the slices preallocated have a fixed length anyway.

For those where we use the rest operator, we at least preallocate the minimum size. The slice will grow automatically beyond that if necessary.

### DIFF
--- a/pkg/reconciler/ingress/lister.go
+++ b/pkg/reconciler/ingress/lister.go
@@ -66,7 +66,6 @@ func (l *gatewayPodTargetLister) ListProbeTargets(ctx context.Context, ing *v1al
 }
 
 func (l *gatewayPodTargetLister) getIngressUrls(ing *v1alpha1.Ingress, gatewayIps []string) ([]status.ProbeTarget, error) {
-	var targets []status.ProbeTarget
 	localDomainName := network.GetClusterDomainName()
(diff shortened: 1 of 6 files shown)


pkg/reconciler/ingress/lister.go
```diff
@@ -66,14 +66,14 @@ func (l *gatewayPodTargetLister) ListProbeTargets(ctx context.Context, ing *v1al
...
 func (l *gatewayPodTargetLister) getIngressUrls(ing *v1alpha1.Ingress, gatewayIps []string) ([]status.ProbeTarget, error) {
-	var targets []status.ProbeTarget
 	localDomainName := network.GetClusterDomainName()
 	ips := sets.NewString()
 
 	for _, ip := range gatewayIps {
 		ips.Insert(ip)
 	}
 
+	targets := make([]status.ProbeTarget, 0, len(ing.Spec.Rules))
```
davidor, line 76: An internal rule generates one target, but an external one generates 2 targets.
The array will grow as needed but maybe we should leave this one as before to avoid confusion?

markusthoemmes: :man_shrugging: we can also add a comment stating that. Why waste precious CPU cycles?

Issue #111, opened 2020-06-28
3scale-kourier-gateway pod fails to start with istio sidecar injection


nak3:
### issue

- When enabled sidecar injection on 3scale-kourier-gateway, the pod is not able to start.
- Even though adding `--base-id ` option, the gateway pod still does not start well.

#### 0. Deploy Serving with Kourier

```
$ cd ${KNATIVE_SERVING_CODE}
$ kubectl apply -f third_party/kourier-latest/kourier.yaml
$ ko apply -f config/  --selector=networking.knative.dev/ingress-provider!=istio
```

#### 1. Deploy Istio

(Please download istioctl 1.6.3 from here https://github.com/istio/istio/releases/tag/1.6.3)

```
$ wget https://gist.githubusercontent.com/nak3/1f533b1a8be1bdd891e9b2fd8bebb40a/raw/d6acf8e0688e046c1c115ab553fd8669a0c78688/istio-mesh.yaml
$ istioctl manifest apply -f istio-ci-mesh.yaml
```

#### 2. Enable sidecar injection on kourier-gateway

```
$ kubectl label namespace kourier-system istio-injection=enabled
$ kubectl  delete pod -n kourier-system --all                   # Restart pods to enable sidecar injection
```

#### 3. Controller can run but gateway does not.

```
$ kubectl  get pod -n kourier-system
NAME                                      READY   STATUS             RESTARTS   AGE
3scale-kourier-control-5485f6df9f-fxb2g   2/2     Running            0          87s
3scale-kourier-gateway-67d8847b79-x2t5k   0/2     CrashLoopBackOff   2          60s

$ kubectl  get pod -n kourier-system
NAME                                      READY   STATUS      RESTARTS   AGE
3scale-kourier-control-5485f6df9f-fxb2g   2/2     Running     0          64s
3scale-kourier-gateway-67d8847b79-x2t5k   0/2     Completed   2          37s
```

#### 4. Adding base-id option

As you can see `unable to bind domain socket with id=0 (see --base-id option)`, the base-id was conflicted.

```
$ kubectl  -n kourier-system logs 3scale-kourier-gateway-67d8847b79-x2t5k -c istio-proxy
  ...
[2020-06-28 07:04:16.886][14][debug][init] [external/envoy/source/common/init/watcher_impl.cc:27] init manager Server destroyed
unable to bind domain socket with id=0 (see --base-id option)
2020-06-28T07:04:16.890205Z	error	Epoch 0 exited with error: exit status 1
[...]

PR #112, opened 2020-06-29
Fixes for kourier to run under istio.

jmprusi:
- Specifies a new base-id for the envoy gateway to avoid clashing with
istio sidecar.
- Specifies missing container ports.
- Name the service ports with the proper protocol..

### DIFF
--- a/deploy/kourier-knative.yaml
+++ b/deploy/kourier-knative.yaml
@@ -111,6 +111,8 @@ spec:
     spec:
       containers:
         - args:
+            - --base-id
+            - "1"
             - -c
             - /tmp/config/envoy-bootstrap.yaml
           command:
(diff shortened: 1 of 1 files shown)


jmprusi:
We can hardcore the base id to 1.. or we can use [use-dynamic-base-id](https://www.envoyproxy.io/docs/envoy/latest/operations/cli#cmdoption-use-dynamic-base-id)

I see `--base-id` is "preferred" but I don't fully understand the implications. 

@nak3 @davidor

jmprusi:
Fixes #111

nak3:
LGTM. I tested on my env and it worked fine. Thank you.
Defer to @davidor for the `/lgtm` label.

PR #117, opened 2020-07-07
Track floating deps @ master

tcnghia:


jmprusi:
@tcnghia  we just merged a PR that unpins de depends and fixes the testing issues: #122 I think we can close this one.

PR #118, opened 2020-07-09
Enables zipkin with sampling set to 0 by default.

jmprusi:
This should be fixed in the next versions of envoy, when we can actually add the custom header x-b3-* directly. (currently, those headers can be manipulated directly) 

With those changes, we can pass the failing test "runtime.TestShouldHaveHeadersSet"

In a next iteration we should add proper tracing support.

### DIFF
--- a/pkg/envoy/http_connection_manager.go
+++ b/pkg/envoy/http_connection_manager.go
@@ -23,9 +23,11 @@ import (
 	accesslog_v2 "github.com/envoyproxy/go-control-plane/envoy/config/accesslog/v2"
 	envoy_accesslog_v2 "github.com/envoyproxy/go-control-plane/envoy/config/filter/accesslog/v2"
 	httpconnectionmanagerv2 "github.com/envoyproxy/go-control-plane/envoy/config/filter/network/http_connection_manager/v2"
+	envoy_type "github.com/envoyproxy/go-control-plane/envoy/type"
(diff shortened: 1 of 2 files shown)


pkg/envoy/http_connection_manager.go
```diff
@@ -54,6 +56,17 @@ func NewHTTPConnectionManager(virtualHosts []*route.VirtualHost) httpconnectionm
 				VirtualHosts: virtualHosts,
 			},
 		},
+		GenerateRequestId: &wrappers.BoolValue{
+			Value: true,
+		},
+		// TODO: Read actual tracing information from knative configmaps
+		// HACK: right now, in order to forward the x-b3-* headers properly, we need to enable zipkin tracing
+		// even if it's set to 0% sampling.
```
nak3, line 64: Is this to pass `TestShouldHaveHeadersSet`? The test was actually passed before with Kourier but it started failing after spoof client dropped  x-b3-* headers by https://github.com/knative/pkg/commit/5658d93fb07ee641a4e0155f3f9c262da7199035.

I am trying to add it back in https://github.com/knative/pkg/pull/1447 :thinking:

jmprusi: That makes more sense! I will close this one then! thanks @nak3

jmprusi:
This is not required, will get fixed by https://github.com/knative/pkg/pull/1447

PR #125, opened 2020-07-15
Add lease RBAC for leader election

tcnghia:


tcnghia:
@mattmoor I backed out the update of deps since it seems to cause TestKourierControlHA to degrade.

tcnghia:
/close

I think there may be some unresolved issues here about the HA test. I'll let @jmprusi / @davidor / @mgencur fix and will try again later.

PR #130, opened 2020-07-22
Add a sleep before stopping the gateway

jmprusi:


nak3:
Hi @jmprusi  I'm sorry to bother you. But may I ask you what this fix is for? Is it related to graceful shutdown or something?
I wanted to know before downstream hit the issue :smile:

jmprusi:
@nak3 no worries! Right now, the lifecycle preStop, just sets the envoy in "failing" mode, and then the pod exits, that's not good, as some requests could still being processed and, any ongoing connection will be killed. By adding a small sleep, we give time to k8s to mark this pod as failing, so it doesn't send any new traffic, and give it some time to finish any ongoing request. 

I'm trying to get all the pending PR's merged by fixing some of the HA flakes.

nak3:
Oh, I understand. Thank you so much @jmprusi for the detail!

PR #132, opened 2020-07-23
Add missing perms for events and coordination via configmap.

jmprusi:


davidor:
The tests are passing in other PRs without these permissions. Is this still needed?

jmprusi:
@davidor yes! Tests are passing, but you will see several errors regarding the lack of permissions for events (not test failures.. ), so this pr fixes a couple things:

* When there's an error/warning, the controller will send an event, right now, that raises an error due to lack of perms.
* If the user has leader-election configure to use configmaps instead of leases, the controller will fail due to lack of perms. 

If you prefer I can split the PR!

PR #136, opened 2020-07-31
Use the readonly interfaces for the new leader election

jmprusi:
This change allows kourier control plane to run in active/active
mode. All PODs will be exposing the configuration but only one (the
leader) will make changes to the k8s knative objects.


tcnghia:
@jmprusi looks like this made the `TestKourierControlHA` test pass  ? can we please merge this change :+1:

PR #140, opened 2020-08-03
Consume conformance tests and test images from `knative.dev/networking`

tcnghia:


tcnghia:
/assign @jmprusi 

There are still some dependencies on `knative.dev/serving` from https://github.com/knative-sandbox/net-kourier/blob/master/test/integration_test.go which is unused.

PR #148, opened 2020-08-11
Update hack/tools.go missed in #140

tcnghia:
/assign @jmprusi 
/assign @mattmoor 

This should fix the auto update deps automation.


tcnghia:
> You probably need to change this too:
> 
> https://github.com/knative-sandbox/net-kourier/blob/35faec6c978a75b7cda85471596e58a679714eef/test/upload-test-images.sh#L32-L36

Done

Issue #155, opened 2020-08-15
Enable Alpha tests


mattmoor:
There are two kingress conformance tests not covered by the main entrypoint by default.

Consider passing `--enable-beta` (nothing yet) and `--enable-alpha` (2 more tests) here: https://github.com/knative-sandbox/net-kourier/blob/fcba41092e623a5adc46d563fbbe6705aadae6fb/test/e2e-tests.sh#L24-L26

mattmoor:
For example: https://github.com/knative-sandbox/net-istio/pull/226

markusthoemmes:
Done!

PR #158, opened 2020-08-18
Enable beta tests

nak3:
This patch enables beta tests.

Part of https://github.com/knative-sandbox/net-kourier/issues/155

The alpha test needs to support RewriteHost so we cannot enable it yet.

/cc @jmprusi @davidor @mattmoor

### DIFF
--- a/test/e2e-tests.sh
+++ b/test/e2e-tests.sh
@@ -23,7 +23,7 @@ failed=0
 
 go_test_e2e -timeout=20m -parallel=12 \
   ./test/conformance \
-  --enable-beta --enable-alpha \
+  --enable-beta \
   --ingressClass=kourier.ingress.networking.knative.dev || failed=1
 
 # Give the controller time to sync with the rest of the system components.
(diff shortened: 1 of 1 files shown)


nak3:
I guess that `TestRewriteHost` does not pass yet, though...

nak3:
@davidor @jmprusi 

Could you pleases take a look? We can enable BETA test which runs [`TestProbeHeaders`](https://github.com/knative/networking/blob/6ae03526c394d2487039e3cad36111682dce99a5/test/conformance/ingress/run.go#L66-L69).

jmprusi:
@nak3 can we disable the hostrewrite test while we implement it?

nak3:
I think we can not... It is an implementation issue of `test.ServingFlags.EnableAlphaFeatures`. 
I will report the issue against https://github.com/knative/networking.

jmprusi:
well this passes without the alpha, so lets merge the beta.

/lgtm

davidor:
It'd be nice if the testing framework provided a way to run those alpha and beta tests in a separate task to make them optional and avoid blocking PRs when they fail.

davidor (approved):
/lgtm

nak3:
Thank you. Yes, agree. I opened https://github.com/knative/networking/issues/183.

dprotaso:
You’ll need to use multiple go test invocations and use -run flag to select the tests you want

PR #164, opened 2020-08-20
Stop expanding cluster local hosts

nak3:
Since knative/serving#9100 Knative expands the cluster local domains in Ingress's spec.rules.hosts.
So Kourier does not needs to expand cluster local domains like:
`hello.myproject.svc.cluster.local` to `hello.myproject.svc`, `hello.myproject`.
This patch stops it.

This patch also refactors:
- to stop bringing cluster local domain but calls `GetClusterDomainName()` when necessary.
- to combine `internalDomain()` with `externalDomain()`.

### DIFF
--- a/pkg/generator/ingress_translator_test.go
+++ b/pkg/generator/ingress_translator_test.go
@@ -176,7 +176,7 @@ func TestIngressVisibility(t *testing.T) {
 		intDomains []string
 		visibility v1alpha1.IngressVisibility
 	}{{
-		name:       "exteranl visibility",
+		name:       "external visibility",
 		hosts:      []string{"hello.default.example.com"},
 		extDomains: []string{"hello.default.example.com", "hello.default.example.com:*"},
 		// External should also be accessible internally
--- a/pkg/generator/ingress_translator_test.go
+++ b/pkg/generator/ingress_translator_test.go
@@ -121,7 +121,7 @@ func TestTrafficSplits(t *testing.T) {
 	}
 
 	ingressTranslator := NewIngressTranslator(
-		kubeClient, newMockedEndpointsLister(), "cluster.local", &pkgtest.FakeTracker{}, logtest.TestLogger(t))
+		kubeClient, newMockedEndpointsLister(), &pkgtest.FakeTracker{}, logtest.TestLogger(t))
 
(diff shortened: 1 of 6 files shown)


nak3:
/retest

Failed some tests that are also failed in the continuous build https://testgrid.knative.dev/net-kourier#continuous :disappointed_relieved:

nak3:
/hold

Since https://github.com/knative/serving/pull/9100 is merged, I need to change the logic.

pkg/reconciler/ingress/lister.go
```diff
@@ -76,8 +73,12 @@ func (l *gatewayPodTargetLister) getIngressUrls(ing *v1alpha1.Ingress, gatewayIp
 	for _, rule := range ing.Spec.Rules {
 		var target status.ProbeTarget
 
-		externalDomains := getExternalDomains(rule, localDomainName)
-		internalDomains := getInternalDomains(rule, localDomainName)
+		var internalDomains, externalDomains []string
+		if rule.Visibility == v1alpha1.IngressVisibilityClusterLocal {
```
davidor, line 77: Do we need this `if`? Both internal and external are the same now.

nak3: Thank you. It was not necessary and we could handle it the following `RuleIsExternal`.

nak3:
@davidor @jmprusi Sorry for bothering you, but could you please take a look again?

pkg/generator/ingress_translator_test.go
```diff
@@ -166,6 +167,66 @@ func TestTrafficSplits(t *testing.T) {
...
+func TestIngressVisibility(t *testing.T) {
+	kubeClient := fake.NewSimpleClientset()
+	tests := []struct {
+		name       string
+		hosts      []string
+		extDomains []string
+		intDomains []string
+		visibility v1alpha1.IngressVisibility
+	}{{
+		name:       "exteranl visibility",
```
jmprusi, line 179: s/exteranl/external

jmprusi (approved):

pkg/generator/ingress_translator_test.go
nak3: Thank you! Fixed.

nak3:
/hold cancel

I forgot that I hold it :cry:

PR #168, opened 2020-09-03
Remove deprecated spec.visibility from KIngress

nak3:
Since https://github.com/knative/networking/commit/f8dacf70c1e8a7daa7dd8847f42f3eb70fc5496a, Kingress deprecates `spec.visibility`.

This patch stops using spec.visibility and use spec.rules.visibility instead.

### DIFF
--- a/pkg/knative/ingress.go
+++ b/pkg/knative/ingress.go
@@ -29,18 +29,8 @@ func MarkIngressReady(ingress *networkingv1alpha1.Ingress) {
 	internalDomain := domainForServiceName(config.InternalServiceName)
 	externalDomain := domainForServiceName(config.ExternalServiceName)
 
-	var domain string
-
-	if ingress.Spec.Visibility == networkingv1alpha1.IngressVisibilityClusterLocal {
-		domain = internalDomain
-	} else {
-		domain = externalDomain
-	}
-
 	ingress.Status.MarkLoadBalancerReady(
-		[]networkingv1alpha1.LoadBalancerIngressStatus{{
-			DomainInternal: domain,
-		}},
+		nil,
 		[]networkingv1alpha1.LoadBalancerIngressStatus{{
 			DomainInternal: externalDomain,
 		}},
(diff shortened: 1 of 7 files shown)


pkg/knative/ingress_rule.go
```diff
@@ -39,15 +39,9 @@ func Domains(rule v1alpha1.IngressRule) []string {
 	return domains
 }
 
-func RuleIsExternal(rule v1alpha1.IngressRule, ingressVisibility v1alpha1.IngressVisibility) bool {
-	switch rule.Visibility {
-	case v1alpha1.IngressVisibilityExternalIP:
-		return true
-	case v1alpha1.IngressVisibilityClusterLocal:
+func RuleIsExternal(rule v1alpha1.IngressRule) bool {
+	if rule.Visibility == v1alpha1.IngressVisibilityClusterLocal {
```
davidor, line 43: If I understood correctly, rule.Visibility is always set now, so this can be simplified to `rule.Visibility == IngressVisibilityExternalIP`

pkg/knative/ingress.go
```diff
@@ -29,18 +29,8 @@ func MarkIngressReady(ingress *networkingv1alpha1.Ingress) {
...
-		domain = internalDomain
-	} else {
-		domain = externalDomain
-	}
-
 	ingress.Status.MarkLoadBalancerReady(
-		[]networkingv1alpha1.LoadBalancerIngressStatus{{
-			DomainInternal: domain,
-		}},
+		nil,
```
davidor, line 33: Why did this change?

pkg/knative/ingress_rule.go
nak3: Yes, that is better. Thank you.

pkg/knative/ingress.go
nak3: Public or Private LB was assigned to `ingress.status.LoadBalancer` based on `ingress.Spec.Visibility` but it is no longer necessary as `ingress.Spec.Visibility`  is deprecated.
I opened the issue against https://github.com/knative/networking/issues/136 to make it deprecate.

nak3:
(The e2e test fixed. I just missed to add visibility in https://github.com/knative-sandbox/net-kourier/pull/168/files#diff-a2dce7ca5a74e1e2bb98f0e7f0d8d346R50.)

davidor (approved):
/lgtm

PR #169, opened 2020-09-03
Preallocate slices, drop unnecessary slice copy, remove deadcode.

markusthoemmes:
As per title.

/assign @jmprusi @davidor

### DIFF
--- a/pkg/knative/ingress_rule.go
+++ b/pkg/knative/ingress_rule.go
@@ -30,9 +30,8 @@ import (
 // The fix is to include ":*" in the domains.
 // This applies both for internal and external domains.
 // More info https://github.com/envoyproxy/envoy/issues/886
-//
 func Domains(rule v1alpha1.IngressRule) []string {
-	var domains []string
+	domains := make([]string, 0, len(rule.Hosts))
 	for _, host := range rule.Hosts {
 		domains = append(domains, host, host+":*")
 	}
--- a/pkg/knative/ingress_rule.go
+++ b/pkg/knative/ingress_rule.go
@@ -31,7 +31,7 @@ import (
 // This applies both for internal and external domains.
 // More info https://github.com/envoyproxy/envoy/issues/886
 func Domains(rule v1alpha1.IngressRule) []string {
-	domains := make([]string, 0, len(rule.Hosts))
+	domains := make([]string, 0, 2*len(rule.Hosts))
 	for _, host := range rule.Hosts {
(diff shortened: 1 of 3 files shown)


davidor (approved):
/lgtm

pkg/reconciler/ingress/lister.go
```diff
@@ -116,9 +116,5 @@ func domainsToURL(domains []string, scheme string) []*url.URL {
 }
 
 func getDomains(rule v1alpha1.IngressRule) []string {
-	var domains []string
-	for _, host := range rule.Hosts {
-		domains = append(domains, host)
-	}
-	return domains
+	return rule.Hosts
```
davidor, line 119: We can get rid of this function and just call `rule.Hosts` directly.

pkg/knative/ingress_rule.go
```diff
@@ -30,9 +30,8 @@ import (
 // The fix is to include ":*" in the domains.
 // This applies both for internal and external domains.
 // More info https://github.com/envoyproxy/envoy/issues/886
-//
 func Domains(rule v1alpha1.IngressRule) []string {
-	var domains []string
+	domains := make([]string, 0, len(rule.Hosts))
```
davidor, line 34: This should be `make([]string, 0, 2*len(rule.Hosts))` , right? Because it's adding 2 domains per host.

pkg/reconciler/ingress/lister.go
```diff
@@ -73,7 +73,7 @@ func (l *gatewayPodTargetLister) getIngressUrls(ing *v1alpha1.Ingress, gatewayIp
 	for _, rule := range ing.Spec.Rules {
 		var target status.ProbeTarget
 
-		domains := getDomains(rule)
+		domains := rule.Hosts
```
nak3, line 76: I am not confidence but it needs to be:

```
hosts := rule.Hosts
domain := hosts
```

I wonder that`domains := rule.Hosts` is problematic like https://golang.org/doc/faq#closures_and_goroutines

markusthoemmes: not sure I parse that. Why would that be problematic :thinking:. If we mutated it, maybe, but then we'd have to clone the entire slice anyway.

nak3: Ah, never mind. The `domains` is created inside the loop and no problem. Sorry for my stupidness :disappointed:

PR #175, opened 2020-09-08
Track master branch again

nak3:
This patch unpinned from v1.7 and start tracking master again.

The master branch deprecated `Retry` (it has been `0` by default for
long time) so this patch removes it as well.


davidor:
For reference, this is the issue that talks about deprecating the `Retries` field: https://github.com/knative/serving/issues/6549

davidor (approved):
/lgtm

PR #178, opened 2020-09-09
Add linting config and fix issues.

markusthoemmes:
As per title. Prep work for knative-sandbox/.github#13 to land in sandbox.

- Moved the ExtAuthz to their own package to unskip them permanently.
- Marked protobuf migration as a TODO.

### DIFF
--- a/.golangci.yaml
+++ b/.golangci.yaml
@@ -1,0 +1,12 @@
+run:
+  timeout: 5m
+
+  build-tags:
+    - e2e
+
+linters:
+  enable:
+    - unconvert
+    - prealloc
+  disable:
+    - errcheck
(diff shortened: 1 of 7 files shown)


davidor:
The external auth test is not run: https://github.com/knative-sandbox/net-kourier/pull/178/checks?check_run_id=1089923528
`?   	knative.dev/net-kourier/test/extauthz	[no test files]`

markusthoemmes:
well, it shouldn't I think? It's an integration test.

davidor:
Ah, you're right. It shouldn't appear there, but it should be in `pull-knative-sandbox-net-kourier-integration-tests`, and it isn't.
Anyway, I think that test will fail because it requires an image that is not built in the test environment.

markusthoemmes:
It was skipped before and I put it in its own package on purpose so we can unskip (and you can run locally easily) but it still isn't picked up by CI. That's how I intended it to work at least :joy:

davidor:
Ah sorry, I misunderstood you. I thought you wanted to run the test in the CI system as well.

davidor (approved):
/lgtm

PR #165, opened 2020-08-26
Add MarkLoadBalancerNotRead when status manager fails to probe

nak3:
When updating an existing Ksvc, Kingress transits to weird status
`NewObservedGenFailure` like this:

```
status:
  conditions:
  - lastTransitionTime: "2020-08-26T02:58:01Z"
    status: "True"
    type: LoadBalancerReady
  - lastTransitionTime: "2020-08-26T02:33:48Z"
    status: "True"
    type: NetworkConfigured
  - lastTransitionTime: "2020-08-26T02:59:50Z"
    message: unsuccessfully observed a new generation
    reason: NewObservedGenFailure
    status: Unknown
    type: Ready
```

This patch adds `MarkLoadBalancerNotRead` when status manager fails to
probe. So the status transits to `Waiting for load balancer to be
ready` like this:

```
status:
  conditions:
  - lastTransitionTime: "2020-08-26T02:58:01Z"
    message: Waiting for load balancer to be ready
    reason: Uninitialized
    status: Unknown
    type: LoadBalancerReady
  - lastTransitionTime: "2020-08-26T02:33:48Z"
    status: "True"
    type: NetworkConfigured
  - lastTransitionTime: "2020-08-26T02:58:01Z"
    message: Waiting for load balancer to be ready
    reason: Uninitialized
    status: Unknown
    type: Ready
```

/cc @jmprusi @davidor

PR #180, opened 2020-09-10
[WIP] Revert "Add MarkLoadBalancerNotRead when status manager fails to probe (#165)

nak3:
This reverts commit f795c3bb5d4eb754b349469ae42b35bdc5c84bf5.

Current continuous build is unstable - https://testgrid.knative.dev/net-kourier#continuous
The ingress status does not become Ready due to `Waiting for load balancer to be ready`.
This patch tries to revert the suspicious commit.


nak3:
`TestKourierControlHA` failed with ingress `timed out waiting for the condition`. The commit is not the issue.

PR #201, opened 2020-09-23
Do not create EnvoyRoute if any of service/endpoint in the Ingress is not ready

nak3:
When endpoint or service was not created, current code does not stop
creating virtualHost but just break the loop.
Due to this, it is possible to create a EnvoyRoute without 100%
weighted_cluster.

For example, when we have the following ingress, net-kourier creates 70% Route if `hello-example-1`
svc is found but `hello-example-2` svc is not found.

```
spec:
  rules:
  - hosts:
    - hello-example.default.18.138.21.218.xip.io
    http:
      paths:
      - splits:
        - appendHeaders:
            Knative-Serving-Namespace: default
            Knative-Serving-Revision: hello-example-1
          percent: 70
          serviceName: hello-example-1
          serviceNamespace: default
          servicePort: 80
        - appendHeaders:
            Knative-Serving-Namespace: default
            Knative-Serving-Revision: hello-example-2
          percent: 30
          serviceName: hello-example-2
          serviceNamespace: default
          servicePort: 80
```

To fix it, this patch returns nil if any of service/endpoint in the
Ingress is not ready.

### DIFF
--- a/pkg/generator/ingress_translator.go
+++ b/pkg/generator/ingress_translator.go
@@ -29,7 +29,6 @@ import (
 	route "github.com/envoyproxy/go-control-plane/envoy/api/v2/route"
 	"go.uber.org/zap"
 	kubev1 "k8s.io/api/core/v1"
-	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	kubeclient "k8s.io/client-go/kubernetes"
 	corev1listers "k8s.io/client-go/listers/core/v1"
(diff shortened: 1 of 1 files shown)


nak3:

Could you please take a look?

The `OnStreamRequest error pushing snapshot to gateway Sum of weights in the weighted_cluster should add up to 100` error stopped showing up in gateway log.

pkg/generator/ingress_translator.go
```diff
@@ -109,15 +109,15 @@ func (translator *IngressTranslator) translateIngress(ctx context.Context, ingre
 				endpoints, err := translator.endpointsLister.Endpoints(split.ServiceNamespace).Get(split.ServiceName)
 				if apierrors.IsNotFound(err) {
 					translator.logger.Warnf("Endpoints '%s/%s' not yet created", split.ServiceNamespace, split.ServiceName)
-					break
+					return nil, nil
```
davidor, line 112: Looks good :+1:
Should we `return nil, err` and delete the warning above ? :thinking:

nak3: Yes, sounds good. Updated.

pkg/generator/ingress_translator.go
```diff
@@ -107,18 +106,12 @@ func (translator *IngressTranslator) translateIngress(ctx context.Context, ingre
 				}
 
 				endpoints, err := translator.endpointsLister.Endpoints(split.ServiceNamespace).Get(split.ServiceName)
-				if apierrors.IsNotFound(err) {
-					translator.logger.Warnf("Endpoints '%s/%s' not yet created", split.ServiceNamespace, split.ServiceName)
-					break
-				} else if err != nil {
+				if err != nil {
```
markusthoemmes, line 109: This will now manifest in immediate reconcile retries, is that what we want? Don't we want to be triggered by endpoints/service events if we can't find something?

nak3: Yes, we want to be triggered by endpoints/services events if it was not found.
I reverted the last change but sorry if I misunderstood it.

davidor (approved):
/lgtm

PR #205, opened 2020-09-24
[Automated] Upgrade to latest dependencies

mattmoor:
Produced via:
```shell
./hack/update-deps.sh --upgrade
./hack/update-codegen.sh
```
/cc davidor markusthoemmes

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -18,7 +18,7 @@ require (
 	k8s.io/client-go v11.0.1-0.20190805182717-6502b5e7b1b5+incompatible
 	knative.dev/networking v0.0.0-20200922180040-a71b40c69b15
 	knative.dev/pkg v0.0.0-20200922164940-4bf40ad82aab
-	knative.dev/serving v0.17.1-0.20200923161440-615c2258f296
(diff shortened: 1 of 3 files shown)


nak3:
/hold

I think current `hack/update-deps.sh` is a little bit wrongly pinned. I will update it.

mattmoor:
@nak3 what do you think is wrong?

nak3:
@mattmoor sorry, never mind.

I found that `serving` is pinned to `master` but others (`pkg`, `networking`, `test-infra`) are `release-0.18`.

https://github.com/knative-sandbox/net-kourier/blob/bf7c4fb3ab51ac1d6cb833bf19808c9d9d049116/hack/update-deps.sh#L34-L39

But I realized that serving does not have the 0.18 branch yet. So, it is by design I think.

/hold cancel

mattmoor:
@nak3 yeah, we're in the weird 1 week period where we're staging pkg before serving cuts.

davidor (approved):
/lgtm

PR #238, opened 2020-10-11
Update deps & use NetworkingFlags instead of ServingFlags

tcnghia:
Part of https://github.com/knative/networking/issues/211 

/assign @nak3


nak3 (approved):
/lgtm

Thank you!

PR #247, opened 2020-10-15
[master] Upgrade to latest dependencies

mattmoor:
Produced via:
```shell
./hack/update-deps.sh --upgrade
./hack/update-codegen.sh
```
/cc davidor markusthoemmes tcnghia
/assign davidor markusthoemmes tcnghia

PR #249, opened 2020-10-19
Stop using test.KubeClient

nak3:
Since https://github.com/knative/pkg/commit/5a5f68637e27295ff78896c290d5d9f4f4c1cead, `test.KubeClient` is not used.

This patch updates it and updates deps like https://github.com/knative-sandbox/net-kourier/pull/247.

/cc @davidor  @jmprusi @tcnghia


tcnghia (approved):
/lgtm

PR #252, opened 2020-10-19
Supply metrics domain via environment

markusthoemmes:
As per title. This aligns Kourier with other Knative componentry.


davidor (approved):
/lgtm

PR #276, opened 2020-10-29
Optimize cache warmup on startup

markusthoemmes:
This makes cache warmup at startup simpler (and most likely faster too):
Instead of relying on the global resync to do its thing (which can be very slow, especially because of our two-lane queues), this runs all the ingresses through the reconciler without queueing them and before launching the config server. Having this seeded config is necessary to start it anyway, so we might as well block startup on it completely.


davidor:
It'd be good to rebase this PR to pick #282 and run the e2e tests.

markusthoemmes:
Done!

PR #282, opened 2020-10-30
Add kind e2e test for github action

nak3:
This patch adds kind e2e test for github action.

It is based on net-contour's https://github.com/knative-sandbox/net-contour/blob/master/.github/workflows/kind-e2e.yaml.

### DIFF
--- a/.github/workflows/kind-e2e.yaml
+++ b/.github/workflows/kind-e2e.yaml
@@ -27,6 +27,10 @@ jobs:
         - v1.18.8
         - v1.19.1
 
+        test-suite:
+        - ./test/conformance
+        - ./test/ha
+
         # Map between K8s and KinD versions.
         # This is attempting to make it a bit clearer what's being tested.
         # See: https://github.com/kubernetes-sigs/kind/releases/tag/v0.9.0
@@ -41,7 +45,6 @@ jobs:
           kind-version: v0.9.0
           kind-image-sha: sha256:98cf5288864662e37115e362b23e4369c8c4a408f99cbc06e58ac30ddc721600
 
-        - test-suite: ./test/conformance
         - test-suite: ./test/ha
           test-flags: "-parallel=1"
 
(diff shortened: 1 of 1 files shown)


.github/workflows/kind-e2e.yaml
```diff
@@ -0,0 +1,190 @@
...
+          sed 's/LoadBalancer/NodePort/g' | \
+          sed 's/imagePullPolicy:/# DISABLED: imagePullPolicy:/g' | \
+          kubectl apply -f -
+
+        # This tells the tests what namespace to look in for our kingress LB.
+        echo "GATEWAY_OVERRIDE=kourier" >> $GITHUB_ENV
+        echo "GATEWAY_NAMESPACE_OVERRIDE=kourier-system" >> $GITHUB_ENV
+
+        # Make sure all pods run in leader-elected mode.
+        kubectl -n kourier-system  scale deployment 3scale-kourier-gateway --replicas=0
```
markusthoemmes, line 140: There's nol leader election in this one I believe. We don't have to scale it down first imo.

.github/workflows/kind-e2e.yaml
```diff
@@ -0,0 +1,190 @@
...
+    - name: Install Knative net-kourier
+      run: |
+        set -o pipefail
+
+        # Build and Publish our containers to the docker daemon (including test assets)
+        export GO111MODULE=on
+        export GOFLAGS=-mod=vendor
+        ko resolve --platform=all -Pf test/config/ -f deploy/kourier-knative.yaml | \
+          sed 's/LoadBalancer/NodePort/g' | \
+          sed 's/imagePullPolicy:/# DISABLED: imagePullPolicy:/g' | \
```
markusthoemmes, line 132: This can go now, I replaced `imagePullPolicy` with `IfNotPresent` everywhere.

davidor:
This is very useful. Thanks :+1:

I have a question regarding the kubernetes versions. I see that this is tested with 1.17.x, 1.18.x and 1.19.x. Is there a policy in knative serving that specifies which kubernetes versions should be supported in each release?

markusthoemmes:
@davidor This is currently under active discussion in the TOC. For now, I think aligning with serving makes the most sense, given that we'll run these tests there potentially too.

davidor (approved):
/lgtm

PR #284, opened 2020-10-30
Update codecov config to not warn on existing coverage

markusthoemmes:
This makes the config equal with serving's and thus is less annoying.


davidor (approved):
/lgtm

PR #285, opened 2020-10-30
Drop dependency overrides and thus bump envoyapi dep

markusthoemmes:
Let's see what shakes out if we try to update our envoyapi deps a bit.


davidor (approved):
/lgtm

PR #286, opened 2020-10-30
Make the update-k8s-dep script consistent with the actual deps

markusthoemmes:
As per title. A small cleanup task.


davidor (approved):
/lgtm

PR #287, opened 2020-10-30
Remove unused make targets and drop duplicated config

markusthoemmes:
Mostly cleanup to get rid of the duplicated bootstrap config.


davidor:
Isn't `docker-build-extauthzutil` still useful? I don't see any other way to generate that image.

markusthoemmes:
@davidor you're somewhat right, but I think all scripts to actually run those tests are currently defunct. We can either add the target back or we setup the tests for the image more like other projects in Knative and fix forward.

PR #288, opened 2020-10-30
Rewrite local setup to use kind instead of k3d

markusthoemmes:
`kind` is widely supported throughout Knative, so Kourier is currently a bit of an oddball. One of the biggest wins is `ko` support, so no `docker` builds are necessary :tada:

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -33,13 +33,13 @@ test: test-unit test-integration test-serving-conformance ## Runs all the tests
 
 test-unit: ## Runs unit tests
 	mkdir -p "$(PROJECT_PATH)/tests_output"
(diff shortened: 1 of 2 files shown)


davidor:
This doesn't work for me, the control pod doesn't start because it says the image name ("ko://knative.dev...") is invalid :confused:

markusthoemmes:
@davidor that sounds like you're on an old `ko` version, can you please check that?

jmprusi:
@davidor I had some issue too.. try to compile ko from `https://github.com/google/ko`

davidor:
Working now. Installing the ko binary was not enough. I had to install `ko` with `go get`. The version was the same (0.6.0) so not sure why it wasn't working.

PR #289, opened 2020-10-30
Rework logging to use more contextualized logging and avoid passing loggers around

markusthoemmes:
As per title, this is a first stab at more principled logging. The context of a reconcile has a logger that is already contextualized, so we can use that instead.


davidor (approved):
/lgtm

PR #290, opened 2020-10-30
Replace panics with proper calls to a logger

markusthoemmes:
As per title.

ExtAuthZ still has some, but replacing them is a bit of a bigger operation, so I'll do that separately.


davidor:
Looks good. Can fix that typo caught by the linter ?

davidor (approved):
/lgtm

PR #292, opened 2020-10-30
Rename common k8s imports to common aliases

markusthoemmes:
Reading the code can be somewhat confusing without these being what they usually are.


davidor (approved):
/lgtm

PR #294, opened 2020-10-30
Remove logger from XDS Server and replace with error handling

markusthoemmes:
Passing the logger into the XDS server seems somewhat unnecessary as we can just log everything in the caller just as well.


davidor:
`finished with error: Failed to create cluster`

/retest

PR #296, opened 2020-10-30
Remove unused CHANGELOG file

markusthoemmes:
Title. This hasn't been used in ages and usually isn't used in Knative releases, apart from the notes on the releases themselves.


davidor:
You're right. It's not useful as it is any more, those are ancient versions. We even changed the version number scheme.

davidor (approved):
/lgtm

PR #297, opened 2020-10-30
Remove (outdated) sample service and integrate into README

markusthoemmes:
As per title. This sample still used the v1alpha1 API which is meanwhile dropped. We can just as well integrate this into the README though.


davidor (approved):
/lgtm

PR #300, opened 2020-11-02
Remove defunct make target for conformance tests

markusthoemmes:
Conformance tests are loaded via the "normal" integration tests anyway. This is defunct as the package in question isn't even vendored.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -7,7 +7,7 @@ local-setup: ## Builds and deploys kourier locally in a k3s cluster with knative
 	./utils/setup.sh
 
 .PHONY: test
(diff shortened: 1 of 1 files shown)


davidor:
I think `test-integration` stopped working. I think it'd be nice to have a makefile that runs both the integration and the conformance tests locally using the cluster created by `make cluster-local`. That was the original idea, but not sure if it's can still be done easily.

markusthoemmes:
@davidor Right, things are fairly broken in that regard. It's on my TODO to fix as well FWIW but feel free to take a crack at it too.

PR #306, opened 2020-11-02
Drop handling of unnecessary ingresses in cache

markusthoemmes:
These seem to serve no real purpose and we have an `Ingress` resource from the reconciler anyway. I don't see a reason to keep tracking them in the cache.

### DIFF
--- a/pkg/reconciler/ingress/ingress.go
+++ b/pkg/reconciler/ingress/ingress.go
@@ -88,14 +88,9 @@ func (r *Reconciler) FinalizeKind(ctx context.Context, ing *v1alpha1.Ingress) re
 }
 func (r *Reconciler) ObserveFinalizeKind(ctx context.Context, ing *v1alpha1.Ingress) reconciler.Event {
 	logger := logging.FromContext(ctx)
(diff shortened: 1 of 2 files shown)


pkg/reconciler/ingress/ingress.go
```diff
@@ -88,14 +88,9 @@ func (r *Reconciler) FinalizeKind(ctx context.Context, ing *v1alpha1.Ingress) re
 }
 func (r *Reconciler) ObserveFinalizeKind(ctx context.Context, ing *v1alpha1.Ingress) reconciler.Event {
 	logger := logging.FromContext(ctx)
-	logger.Infof("Deleting Ingress %s namespace: %s", ing.Name, ing.Namespace)
-	ingress := r.caches.GetIngress(ing.Name, ing.Namespace)
+	logger.Infof("Deleting Ingress %s/%s", ing.Name, ing.Namespace)
```
davidor, line 91: It seems that we were saving the ingresses in the cache because this function didn't receive and instance of ingress before: https://github.com/knative-sandbox/net-kourier/commit/af048dee55924a7b03cffac03c6d5a7d706dbe45

Good to see that we don't need all that anymore :+1:

davidor (approved):
/lgtm

PR #307, opened 2020-11-02
Cleanup some cache helpers and remove indirections

markusthoemmes:
Indirections make stuff hard to follow and in this case, those seem unnecessary. So let's drop them.

### DIFF
--- a/pkg/generator/caches.go
+++ b/pkg/generator/caches.go
@@ -52,8 +52,17 @@ func NewCaches(ctx context.Context, logger *zap.SugaredLogger, kubernetesClient 
 		clusters:            newClustersCache(logger.Named("cluster-cache")),
 		clustersToIngress:   make(map[string][]string),
 		logger:              logger,
+		statusVirtualHost:   statusVHost(),
 	}
-	return c, c.initConfig(ctx, kubernetesClient, extAuthz)
+
+	if extAuthz {
+		c.addClusterForIngress(config.ExternalAuthz.Cluster, "__extAuthZCluster", "_internal")
+	}
+
+	if err := c.setListeners(ctx, kubernetesClient); err != nil {
+		return nil, err
+	}
+	return c, nil
 }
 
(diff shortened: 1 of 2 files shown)


pkg/generator/caches_test.go
```diff
@@ -187,7 +187,8 @@ func createTestDataForIngress(
 		},
 	}, &translatedIngress)
 
-	caches.addStatusVirtualHost()
+	statusVirtualHost := statusVHost()
+	caches.statusVirtualHost = &statusVirtualHost
```
davidor, line 191: Is this needed? Should be already done in this call, right?
https://github.com/knative-sandbox/net-kourier/pull/307/files#diff-99be96ee4d2233307511fe08afab35fce19f6bb870f9a92d140cee81a2fa3b4aR44

markusthoemmes: Good call, fixed!

pkg/generator/caches.go
```diff
@@ -47,13 +47,23 @@ type Caches struct {
 }
 
 func NewCaches(ctx context.Context, logger *zap.SugaredLogger, kubernetesClient kubeclient.Interface, extAuthz bool) (*Caches, error) {
+	statusVirtualHost := statusVHost()
 	c := &Caches{
 		translatedIngresses: make(map[string]*translatedIngress),
 		clusters:            newClustersCache(logger.Named("cluster-cache")),
 		clustersToIngress:   make(map[string][]string),
 		logger:              logger,
+		statusVirtualHost:   &statusVirtualHost,
```
davidor, line 56: Given that it's not used anywhere else, we could do `&statusVHost()` here and delete the assignment above.

markusthoemmes: Nope, that notation doesn't work. If we merge #309 first though, we can do that.

davidor (approved):
/lgtm

PR #308, opened 2020-11-02
Pass nil maps instead of 0-len maps to avoid a few allocations

markusthoemmes:
Likely doesn't really matter in the grand scheme of things, but a good practice to follow nonetheless!


davidor (approved):
/lgtm

PR #309, opened 2020-11-02
Make VirtualHost generators return pointers directly

markusthoemmes:
This seems to be just about the only resource that's not generated as a pointer from the get-go, but it's used as a pointer everywhere. So let's spare ourselves the intermediate variables.


davidor (approved):
/lgtm

PR #311, opened 2020-11-02
Remove unnecessary ingress -> cluster mapping

markusthoemmes:
This mapping doesn't seem to serve any purpose currently, so why not drop the mapping entirely?


davidor (approved):
/lgtm

PR #312, opened 2020-11-02
Drop all usages of Knative types on caching layer

markusthoemmes:
This makes the split of the translation -> cache -> snapshot layers a bit more apparent by not having any of the initial types in the caching layer at all.

### DIFF
--- a/pkg/generator/caches.go
+++ b/pkg/generator/caches.go
@@ -32,7 +32,6 @@ import (
 	kubeclient "k8s.io/client-go/kubernetes"
 	"knative.dev/net-kourier/pkg/config"
 	"knative.dev/net-kourier/pkg/envoy"
-	"knative.dev/networking/pkg/apis/networking/v1alpha1"
 )
 
 var ErrDomainConflict = errors.New("ingress has a conflicting domain with another ingress")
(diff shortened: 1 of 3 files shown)


pkg/generator/caches.go
```diff
@@ -92,8 +91,8 @@ func (caches *Caches) validateIngress(translatedIngress *translatedIngress) erro
 	return nil
 }
 
-func (caches *Caches) addTranslatedIngress(ingress *v1alpha1.Ingress, translatedIngress *translatedIngress) error {
-	caches.logger.Debugf("adding ingress: %s/%s", ingress.Name, ingress.Namespace)
+func (caches *Caches) addTranslatedIngress(translatedIngress *translatedIngress) error {
+	caches.logger.Debugf("adding ingress: %s/%s", translatedIngress.name.Name, translatedIngress.name.Namespace)
```
davidor, line 95: More related to the previous PR than this one, but maybe we should add a `String()` function in `NamespacedName`?

pkg/generator/caches.go
```diff
@@ -106,7 +105,7 @@ func (caches *Caches) addTranslatedIngress(ingress *v1alpha1.Ingress, translated
 	caches.translatedIngresses[translatedIngress.name] = translatedIngress
 
 	for _, cluster := range translatedIngress.clusters {
-		caches.clusters.set(cluster, ingress.Name, ingress.Namespace)
+		caches.clusters.set(cluster, translatedIngress.name.Name, translatedIngress.name.Namespace)
```
davidor, line 108: Should this function accept a `NamespacedName` instead of a name and a namespace? Same for the others like `caches.deleteTranslatedIngress()`.

pkg/generator/caches.go
markusthoemmes: oops, I could've used this here. It has that. Lemme do a pass to turn all namespace/name usages into the right direction too :joy:

PR #317, opened 2020-11-03
Remove unnecessary YAML configuration from test/config

markusthoemmes:
All of the configs and RBAC stuff shouldn't be needed as nothing is referencing it.


davidor (approved):
/lgtm

PR #318, opened 2020-11-03
Fix the 'test-integration' target of the Makefile to successfully run tests locally

markusthoemmes:
This does a few things to successfully bring back local integration testing:

1. Replace port-forwarding with setting up a NodePort.
2. Add a script geared towards local testing that fetches the correct IP etc.
3. Call that script.

**Note:** This doesn't contain extAuth tests just yet. They need a bit more massaging.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -18,7 +18,7 @@ test-unit-coverage: ## Runs unit tests and generates a coverage report
 	go tool cover -html="$(PROJECT_PATH)/tests_output/unit.cov"
 
 test-integration: local-setup ## Runs integration tests
-	go test -mod vendor -race test/*.go
+	./test/e2e-kind.sh
 
 .PHONY: fmt
 fmt: ## Runs code formatting
--- a/test/e2e-kind.sh
+++ b/test/e2e-kind.sh
@@ -18,6 +18,7 @@
 
 set -euo pipefail
 
+export KO_DOCKER_REPO=kind.local
 $(dirname $0)/upload-test-images.sh
 
 ips=( $(kubectl get nodes -lkubernetes.io/hostname!=kind-control-plane -ojsonpath='{.items[*].status.addresses[?(@.type=="InternalIP")].address}') )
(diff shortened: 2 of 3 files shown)


jmprusi:
KO_DOCKER_REPO is unset.. is that expected?

markusthoemmes:
Good point, likely worked for me locally cause I have it set globally, fixing.

jmprusi:
As expected.. macOS & docker doesn't play too well with this.

davidor:
I got an error while publishing the images:
```bash
./test/e2e-kind.sh
>> Publishing test images
2020/11/03 12:25:12 Using base gcr.io/distroless/static:nonroot for knative.dev/net-kourier/vendor/knative.dev/networking/test/test_images/grpc-ping
2020/11/03 12:25:12 No matching credentials were found, falling back on anonymous
2020/11/03 12:25:13 Building knative.dev/net-kourier/vendor/knative.dev/networking/test/test_images/grpc-ping for linux/amd64
2020/11/03 12:25:17 Loading kind.local/grpc-ping:afb39c07a7b4687e62dd25bd697b8a2d98afca5b3bf62f3237d40976249c4c92
2020/11/03 12:25:17 error processing import paths in "-": error resolving image references: no nodes found for cluster "kind"
```

markusthoemmes:
@jmprusi ohhhh... good point. Using the port-forwarding would fix that, right?
@davidor good one, pushed another fix.

jmprusi:
@markusthoemmes yes, we can leave the node Port, but we should put some "disclaimer" there: "if you are running this on macOS use port-forward" or similar

markusthoemmes:
I'll try to change the script to use port-forward. I'd like to make this as easy as possible for everybody.

markusthoemmes:
Actually @jmprusi it'd be great if we could move forward with this at is and we look at supporting macOS separately if possible. Given that we're completely defunct today anyway, this is at least an improvement.

PR #319, opened 2020-11-03
Simplify unit test targets a bit

markusthoemmes:
We don't actually need to be "smart" with the packages, as the integration tests all carry build tags. `-mod vendor` is automatically added by go 1.14 (which is required by the repo).

This also pulls apart `test-unit` from `test-unit-coverage`.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -12,7 +12,7 @@ test: test-unit test-integration ## Runs all the tests
 test-unit: ## Runs unit tests
 	go test -race ./...
 
-test-unit-coverage: test-unit ## Runs unit tests and generates a coverage report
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -10,15 +10,16 @@ local-setup: ## Builds and deploys kourier locally in a "kind" cluster with knat
 test: test-unit test-integration ## Runs all the tests
 
 test-unit: ## Runs unit tests
+	go test -race ./...
+
+test-unit-coverage: test-unit ## Runs unit tests and generates a coverage report
```
davidor, line 15: This runs the tests twice

markusthoemmes: ack!

davidor (approved):
/lgtm

PR #320, opened 2020-11-03
Use 'auto' for coverage checks to rely on the current actual coverage

markusthoemmes:
We've never even been close to 80% (yet). We should revisit this later but get it out of the way for the time being.


davidor (approved):
/lgtm

PR #321, opened 2020-11-03
Split config into files and rename the folder to config

markusthoemmes:
As per title, this aligns us closer with other Knative projects in that we use the canonical `config` folder and split the YAML files into logical components to make changing and reading them somewhat easier.

### DIFF
--- a/.github/workflows/kind-e2e.yaml
+++ b/.github/workflows/kind-e2e.yaml
@@ -127,7 +127,7 @@ jobs:
         # Build and Publish our containers to the docker daemon (including test assets)
         export GO111MODULE=on
         export GOFLAGS=-mod=vendor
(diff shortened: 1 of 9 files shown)


davidor:
What do the prefixes (100, 200, 300-) of the yaml files mean?

markusthoemmes:
They loosely enforce an ordering, especially for the release files.

davidor:
I don't have a strong preference about this. @jmprusi do you see any problem in splitting the yamls?
Also, @markusthoemmes I don't know if we'll need to change something in the serving repo to pick the new yamls. Not sure how that's done now.

markusthoemmes:
Serving only consumes the releases generated from this repo (including nightly releases), so by me adjusting `release.sh` we should be fine wrt. others consuming this as their interface doesn't change.

PR #322, opened 2020-11-03
Split envoy package into API and Server

markusthoemmes:
As per title, to have a clear cut between the API helpers we create and the XDS Server and utilities.


davidor (approved):
/lgtm

Issue #105, opened 2020-06-17
Kourier control is not updated when CERTS_SECRET_NAME is updated


StarpTech:
I had to restart the 3scale-kourier-control.

jmprusi:
Hi @StarpTech ! That shouldn't happen! Can you help us by describing a little bit more what did you do?

StarpTech:
Sry for the incomplete bug report. As far as I remember, after following the https://github.com/knative/net-kourier/blob/05d056b1c9cc601dc2f19d882078014ec28cdbcb/README.md#setup-tls-certificate steps my cert was not set. Only after `kubectl rollout restart deploy ...` the cert was found.

markusthoemmes:
Hey @StarpTech. I'm currently in the process of reworking Kourier a bit and was wondering if you'd be up for a chat on how we can make this feature a bit more... bulletproof.

PR #323, opened 2020-11-03
Remove functionality to front everything with a global cert

markusthoemmes:
This might be a bit controversial and we also might want to open at least an issue to bring this functionality back. I mostly removed it here to simplify the paths where kubeclients are passed around. As can be seen in the diff, this constraints Kubernetes client usage to the `ingress -> translatedIngress` layer and thus makes it fairly straightforward to reason about.

### DIFF
--- a/pkg/generator/caches.go
+++ b/pkg/generator/caches.go
@@ -17,7 +17,6 @@ limitations under the License.
 package generator
 
 import (
-	"context"
 	"errors"
 	"sync"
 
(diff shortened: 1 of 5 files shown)


davidor:
Not sure about this one. There was an issue about this functionality https://github.com/knative-sandbox/net-kourier/issues/105, so we know that at least it was interesting for somebody.

markusthoemmes:
Interesting, thanks @davidor! I'll try to reach out there and see if we can find a sweetspot that doesn't rely on fetching the secret everytime as I doubt this currently scales at all.

markusthoemmes:
Closing this for now.

PR #324, opened 2020-11-03
Use correct name of example service in "utils/setup.sh"

davidor:
Changes the example curl command to use the same service name as the one shown in the README.


markusthoemmes (approved):
/lgtm
/approve

PR #325, opened 2020-11-03
Move code to where it is used and remove 'knative' package

markusthoemmes:
IMO the Knative package serves no real purpose, so I moved some helpers into their respective callsites and some into `config`, where they imo belong.


pkg/config/config.go
```diff
@@ -34,3 +41,19 @@ const (
...
 	KourierIngressClassName = "kourier.ingress.networking.knative.dev"
 )
+
+// ServiceDomains returns the external and internal service's respective domain.
+//
+// Example: kourier.kourier-system.svc.cluster.local.
+func ServiceDomains() (string, string) {
+	suffix := "." + GatewayNamespace() + ".svc." + network.GetClusterDomainName()
+	return ExternalServiceName + suffix, InternalServiceName + suffix
+}
```
tcnghia, line 51: can we use https://github.com/knative/pkg/blob/c46a18644992a3ead940f96f7b9e9078c0dcbb28/network/domain.go#L39

markusthoemmes: Thanks, yes we can. Pushed!

davidor:
I'm not sure about this one. From the point of view of Kourier, ideally, the functions that only accept knative entities as params, like an `Ingress` or and `IngressRule` should be provided in the knative libs. Not all of them might be general enough to be included in the knative libs, though. When they're not provided, we can place them in our own "knative" package. For me, the purpose of the package is clear: provide helpers that we'd like to have in the knative libs but are not there.

markusthoemmes:
I disagree @davidor. The helpers are all quite Kourier specific. The `Domains` helper for example, adds stuff that is specific to envoy's gRPC handling.

I'd especially like to move them because we only use most of them just once, so there's super little value in having a potential (non-straightforward) indirection in there. The `MakeIngressReady` helper is such an instance for example, where it's just clearer to have it inline rather than have it in an indirection.

davidor:
In the case you mentioned, you're right. It's specific to Kourier because it's something that's needed because of the way Envoy handles those domains. On the other hand, I think it'd be nice to have an `IsExternal` function in `IngressRule` instead of having to match constants. Anyway, it's just a function and if we inline it we can delete a package after this refactor. Let's merge :+1:

davidor (approved):
/lgtm

PR #329, opened 2020-11-04
Use 'global' client instead of passing it through in each request

markusthoemmes:
This simplifies calling code a bit by not having to pass kubeclients through at so many occasions. Instead, the cache takes a kubeclient (as it used to anyway) and uses that in its methods.


davidor (approved):
/lgtm

PR #330, opened 2020-11-04
Avoid certificate conversion to string from secret

markusthoemmes:
A secret delivers `[]byte` and the API requires `[]byte`, so there is no reason to cast in between. Casting into a string is relatively costly and not necessary at all in this case as the value is opaque anyway.


davidor (approved):
/lgtm

PR #332, opened 2020-11-04
Reduce unnecessary RBAC rules

markusthoemmes:
I don't think we ever update or create ConfigMaps, nor do we need to do anything with CRDs or Namespaces, I believe.

### DIFF
--- a/config/200-serviceaccount.yaml
+++ b/config/200-serviceaccount.yaml
@@ -30,13 +30,13 @@ metadata:
 rules:
   - apiGroups: [""]
     resources: ["events"]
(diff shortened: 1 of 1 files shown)


davidor:
Namespaces and configmaps were added to avoid some errors/warnings that don't make the tests fail: https://github.com/knative-sandbox/net-kourier/pull/132
Not sure if that has changed.

markusthoemmes:
Hmm, that added events and wide authorization for configmaps. I'd think that if tests pass, this is fine.

tcnghia (approved):
/lgtm

PR #336, opened 2020-11-04
Collapse slices, collapse return values and remove whitespace

markusthoemmes:
This is strictly a formatting change, so everything is consistent with each other.


davidor:
I don't mind the change, but it will be difficult to keep it always like this if it's not enforced by one of the github action checks.

markusthoemmes:
Sadly, I have yet to find a formatter that does this for us :(

davidor (approved):
/lgtm

PR #338, opened 2020-11-05
Do not reconcile endpoints if the ready addresses haven't changed

markusthoemmes:
This should significantly reduce churn produced by endpoint reconciles as we'll no longer reconcile if non ready endpoints are added for example.

### DIFF
--- a/pkg/reconciler/ingress/controller.go
+++ b/pkg/reconciler/ingress/controller.go
@@ -24,6 +24,7 @@ import (
 	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/apimachinery/pkg/util/sets"
 	"k8s.io/client-go/tools/cache"
 	"knative.dev/net-kourier/pkg/config"
 	envoy "knative.dev/net-kourier/pkg/envoy/server"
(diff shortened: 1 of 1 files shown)


pkg/reconciler/ingress/controller.go
```diff
@@ -238,3 +252,23 @@ func getReadyIngresses(ctx context.Context, knativeClient networkingClientSet.Ne
...
+	var count int
+	for _, subset := range eps.Subsets {
+		count += len(subset.Addresses)
+	}
+
+	if count == 0 {
+		return nil
+	}
+
+	ready := make(sets.String, count)
```
davidor, line 266: I wonder if it's worth it to calculate `count` in this case.

markusthoemmes: It should be super quick because endpoints usually have at most a handful of subsets (which is what we need to iterate over to compute it).

davidor: Yeah, in this case, I would personally just delete those lines unless I measured that this has a significant perf impact.
But it's just personal preference :+1:

davidor (approved):
/lgtm

pkg/reconciler/ingress/controller.go
markusthoemmes: I agree with that sentiment in principle. I followed what we do in other similar places here. I'll write benchmarks for some helpers upstream though, to proof/disproof this and then adjust.

PR #346, opened 2020-11-10
Remove unused NewRouteStatusOK

nak3:
This patch removes unused NewRouteStatusOK which creates a route to return 200.
It was not used since e9e16bcbc60b1b3771320164c6606e007576b2cd.


davidor (approved):
/lgtm

PR #347, opened 2020-11-10
Fix cluster-cache test to be correct and quicker

markusthoemmes:
The cluster cache test currently does not test what we actually want to test. It instead uses a parameter that will never be set in production.

This fixes that and also makes the test considerably quicker. It used to take 2 seconds for no apparent reason.

/assign @jmprusi @davidor

### DIFF
--- a/pkg/generator/cluster_cache.go
+++ b/pkg/generator/cluster_cache.go
@@ -66,7 +66,7 @@ func newClustersCacheWithExpAndCleanupIntervals(expiration time.Duration, cleanu
 
 func (cc *ClustersCache) set(cluster *v2.Cluster, ingressName string, ingressNamespace string) {
 	key := key(cluster.Name, ingressName, ingressNamespace)
-	cc.clusters.Set(key, cluster, gocache.DefaultExpiration)
+	cc.clusters.Set(key, cluster, gocache.NoExpiration)
 }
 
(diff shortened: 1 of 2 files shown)


pkg/generator/cluster_cache.go
```diff
@@ -46,33 +46,33 @@ import (
...
-	goCache := gocache.New(expiration, cleanupInterval)
-	return &ClustersCache{clusters: goCache}
+	goCache := gocache.New(gocache.NoExpiration, cleanupInterval)
+	return &ClustersCache{clusters: goCache, clusterExpiration: expiration}
 }
 
 func (cc *ClustersCache) set(cluster *v2.Cluster, ingressName string, ingressNamespace string) {
 	key := key(cluster.Name, ingressName, ingressNamespace)
-	cc.clusters.Set(key, cluster, defaultExpiration)
+	cc.clusters.Set(key, cluster, gocache.DefaultExpiration)
```
davidor, line 69: They're equivalent but I think that using `gocache.NoExpiration` here expresses better what we want to achieve, which is don't expire anything unless we call `setExpiration`.

markusthoemmes: ack, done.

davidor (approved):
/lgtm

PR #348, opened 2020-11-10
Cleanup tests by using proper assertions etc

markusthoemmes:
The assertion library used is quite powerful in what helpful output it can give. So let's make use of that!

This also cleans up a bit of noop setup etc.

/assign @jmprusi @davidor

### DIFF
--- a/pkg/generator/caches_test.go
+++ b/pkg/generator/caches_test.go
@@ -38,9 +38,7 @@ func TestDeleteIngressInfo(t *testing.T) {
 	ctx := context.Background()
 
 	caches, err := NewCaches(ctx, &kubeClient, false)
-	if err != nil {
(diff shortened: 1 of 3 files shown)


pkg/generator/cluster_cache.go
```diff
@@ -46,33 +46,33 @@ import (
 )
 
 const (
-	clusterExpiration      = 15 * time.Second
-	defaultExpiration      = gocache.NoExpiration
+	defaultExpiration      = 15 * time.Second
```
davidor, line 49: These are changes from the previous PR.

markusthoemmes: Indeed, lemme undo.

davidor (approved):
/lgtm

PR #351, opened 2020-11-11
Avoid setting the envoy snapshot twice per reconcile

markusthoemmes:
This **seems** unnecessary and maybe was needed because we used older envoy versions. Let's see what happens if we don't do this.

/assign @jmprusi @davidor

### DIFF
--- a/pkg/reconciler/ingress/ingress.go
+++ b/pkg/reconciler/ingress/ingress.go
@@ -21,7 +21,6 @@ import (
 	"errors"
 	"fmt"
 
-	"github.com/envoyproxy/go-control-plane/pkg/cache"
 	"knative.dev/net-kourier/pkg/config"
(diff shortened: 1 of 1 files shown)


markusthoemmes:
Running here for additional stability testing: https://github.com/knative/serving/pull/10091

davidor:
The PR that added this change is https://github.com/knative-sandbox/net-kourier/pull/45
There's a discussion there about why this was needed. I guess this was fixed in a recent version of Envoy. If we don't see any tests failing with the error in that PR I think it should be good to merge :+1:

jmprusi:
Some of the serving tests failed, not sure if a flake... https://github.com/knative/serving/pull/10091

@markusthoemmes can you rebase and retry those tests?

markusthoemmes:
@jmprusi done and all passed.

PR #356, opened 2020-11-17
Drop timed caching of clusters

markusthoemmes:
# Changes

Tries dropping the cluster caching to keep clusters longer in config than they are actually needed.


markusthoemmes:
Okay, this doesn't "just" work sadly. Imma look into it though.

PR #357, opened 2020-11-17
Cleanup functions around getting/setting snapshots

markusthoemmes:
- `GetSnapshot` isn't used anymore at all.
- `SetSnapshot` doesn't have to take a pointer just to dereference it again to pass it on.
- Copied the order from envoy types.

/assign @jmprusi @davidor


davidor (approved):
/lgtm

PR #358, opened 2020-11-17
Bump envoy version to 1.14.4

markusthoemmes:
As per title. This bumps the image to Maistra's 2.0.0 image, which has envoy 1.14.4.

The regex configuration got dropped and replaced as shown.

/assign @jmprusi @davidor


davidor (approved):
/lgtm

PR #362, opened 2020-11-18
Fix all deprecation warnings from Envoy

markusthoemmes:
Further future proofing for upcoming Envoy updates. This fixes all current deprecation warnings we're getting from Envoy.

/assign @jmprusi @davidor


davidor (approved):
/lgtm

PR #364, opened 2020-11-18
Use IDHasher from envoy's library

markusthoemmes:
As per title, the implementation is roughly the same and should be just fine for our matters.

/assign @jmprusi @davidor


davidor (approved):
/lgtm

PR #367, opened 2020-11-18
Don't even try to probe services that have a domain conflict

markusthoemmes:
During translation to the new reconciler framework, we lost the ability to surface domain conflicts from the reconciler because we're not bailing out of reconcilation before we reach the prober, which will reset to `MarkLoadBalancerNotReady`. This fixes that.

This also drops the loglevel of the conflict log to info, as it's not really an error from the reconciler's PoV.

/assign @jmprusi @davidor

### DIFF
--- a/pkg/reconciler/ingress/ingress.go
+++ b/pkg/reconciler/ingress/ingress.go
@@ -51,8 +51,8 @@ func (r *Reconciler) ReconcileKind(ctx context.Context, ing *v1alpha1.Ingress) r
 	if err := r.updateIngress(ctx, ing); errors.Is(err, generator.ErrDomainConflict) {
 		// If we had an error due to a duplicated domain, we must mark the ingress as failed with a
 		// custom status. We don't want to return an error in this case as we want to update its status.
-		logging.FromContext(ctx).Info("Ingress rejected as its domain conflicts with another ingress")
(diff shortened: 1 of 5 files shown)


pkg/reconciler/ingress/ingress.go
```diff
@@ -45,9 +45,18 @@ var _ ingress.Finalizer = (*Reconciler)(nil)
...
 
 func (r *Reconciler) ReconcileKind(ctx context.Context, ing *v1alpha1.Ingress) reconciler.Event {
+	ing.SetDefaults(ctx)
 	before := ing.DeepCopy()
 
-	r.ObserveKind(ctx, ing)
+	if err := r.updateIngress(ctx, ing); errors.Is(err, generator.ErrDomainConflict) {
+		// If we had an error due to a duplicated domain, we must mark the ingress as failed with a
+		// custom status. We don't want to return an error in this case as we want to update its status.
+		logging.FromContext(ctx).Info("Ingress rejected as its domain conflicts with another ingress")
```
davidor, line 54: The error message is repeated 3 times in this file. Would be good to extract it to a constant.

markusthoemmes: Done.

davidor (approved):
/lgtm

PR #371, opened 2020-11-19
Name envoy config resources in a non-conflicting way

markusthoemmes:
Newer envoy versions seem to be more sensitive to how resources are named. Especially the VirtualHosts started colliding when multiple `Rules` are at play (which is often the case in Serving).

/hold

I want to send the test into the conformance suite and test this through serving thoroughly first too.


markusthoemmes:
Verified to work in https://github.com/knative/serving/pull/10151

markusthoemmes:
/unhold

Ready to go forward.

davidor (approved):

PR #377, opened 2020-11-23
Setup an upgrade test to catch upgrade regressions

markusthoemmes:
As per title. To make sure we don't regress in our upgradeability.


markusthoemmes:
/assign @jmprusi @davidor 

This works thanks to the 0.19.1 upgrade now.

PR #381, opened 2020-11-24
[RELEASE-0.19] Add an upgrade test

markusthoemmes:
As per title, this adds an upgrade test to catch upgradeability regressions and to make sure we can **always** upgrade from the last version to the new one.


markusthoemmes:
/assign @jmprusi @davidor 

0.18 -> 0.19 upgrade is safe!

PR #385, opened 2020-11-26
Fix max port comparison

markusthoemmes:
The current code leads to a build failure during release on 32 bit architectures

```
pkg/config/ext_authz.go:78:21: constant 4294967295 overflows int
```

The code was wrong anyway.

/assign @jmprusi @davidor


davidor (approved):
/lgtm

PR #412, opened 2021-01-03
Remove unused EnvoyNodeID value

nak3:
This patch makes a tiny change which removes unused `EnvoyNodeID` value.

The node ID is defined as `nodeID` in [controller.go](https://github.com/knative-sandbox/net-kourier/blob/42e7f0fbf7337c4d50af790b2ef9dad4a6557026/pkg/reconciler/ingress/controller.go#L56).

# Changes

/kind cleanup

**Release Note**

```release-note
NONE
```

### DIFF
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -26,8 +26,6 @@ import (
 const (
 	ControllerName = "kourier"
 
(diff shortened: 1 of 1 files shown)


davidor (approved):
/lgtm

PR #431, opened 2021-02-02
Add e2e test with newer Envoy image

nak3:
This patch adds the latest envoy image in e2e test.

Maistra is a fork of Envoy but there is no published multi-arch build
and the version is behind envoy upstream so it should be worth to add
the test.
In fact, the new test detected that Envoy 1.17+ does not work yet.

/cc @davidor @jmprusi @markusthoemmes


config/envoy/300-gateway.yaml
```diff
@@ -0,0 +1,205 @@
+# Copyright 2020 The Knative Authors
```
markusthoemmes, line 1: We should strive to keep the diff between the two files at 0, modulo the image reference. Is that not the case right now?

nak3: Yeah, it would be better to keep the no diff. Updated. (I hope I didn't misunderstand your comment.)

markusthoemmes (approved):
/lgtm
/approve

Awesome! Thanks!

PR #433, opened 2021-02-02
Use generic envoy image by default but keep testing Maistra

markusthoemmes:
This makes "upstream" more generic while still making sure that net-kourier still works fine with Maistra's images.


evankanderson (commented):
/lgtm
/approve

I think there may be an 1.17 Envoy coming as well...

markusthoemmes:
@evankanderson we're working on that. See the API v3 PR.

PR #440, opened 2021-02-08
Bump a few assorted dependencies to their latest versions

markusthoemmes:
As per title, just to keep us chasing the latest and greatests version. A bit of a monday cleanup!

Also drops a bit of the unnecessary error handling around UUID generation.


davidor (approved):
/lgtm

PR #442, opened 2021-02-10
set alpn protocols to h2,http/1.1 on downstream tls context

NickCao:
# Changes
- :broom: Set alpn protocols on downstream tls context, in order to leverage h2.

/kind enhancement

### DIFF
--- a/pkg/envoy/api/listener.go
+++ b/pkg/envoy/api/listener.go
@@ -170,9 +170,6 @@ func createTLSContext(certificate []byte, privateKey []byte) *auth.DownstreamTls
 	return &auth.DownstreamTlsContext{
 		CommonTlsContext: &auth.CommonTlsContext{
 			AlpnProtocols: []string{"h2", "http/1.1"},
-			//TlsParams: &auth.TlsParameters{
-			//	TlsMinimumProtocolVersion: auth.TlsParameters_TLSv1_2,
-			//},
 			TlsCertificates: []*auth.TlsCertificate{{
 				CertificateChain: &core.DataSource{
 					Specifier: &core.DataSource_InlineBytes{InlineBytes: certificate},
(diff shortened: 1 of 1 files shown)


NickCao:
@googlebot I signed it!

markusthoemmes:
/ok-to-test

Thanks for this! Could you comment on the implications that this might have?

NickCao:
As far as I known, tls alpn is used for the selection of application layer protocol atop tls. And in the negotiation process, server offers a list of acceptable protocols, from which the client choose one. When no such list is offered, clients tend to choose safe defaults, aka http/1.1, which has serious performance impacts in modern day environments, Thus we should set alpn protocols on downstream tls context. And for the reason why http/1.1 and http/2 is used, it's more of a defacto standard, I checked the behavior of other popular ingress controllers, such as kong, and they all offered these values.

NickCao:
While I'm confident about the consequences of setting the alpn protocols, the minimum tls protocol version is more of a personal preference, and it should be more appropriately exposed as a config value, thus I comment it out to serve as a reminder.

markusthoemmes:
Sounds good to me

/assign @nak3 @jmprusi @davidor 

For more expertise.

Thanks for the contribution and sorry for having it dangling for so long!

nak3:
>  And (possibly) set tls minimum version to tls 1.2, as tls 1.0 and 1.1 should be deprecated for security.

Yes, I think it is better to set the TLS minimum version to 1.2 as 1.1 is EOL. 
But we should discuss it in separated issue/PR. Could you delete the comment about tls version from https://github.com/knative-sandbox/net-kourier/pull/442#issue-570736563 ? 

For alpn change, it looks good to me.

NickCao:
All set, I also removed the comments in the code to avoid confusions.

markusthoemmes (approved):
Thanks!

PR #447, opened 2021-02-13
Split net-kourier configmap and add placeholder.go

mattmoor:
Split the bootstrap configmap into its own file and add a placeholder go file to enable vendoring net-kourier's configuration via `go mod vendor`.

see: https://github.com/mattmoor/mink/pull/449/files


davidor (approved):
/lgtm

PR #457, opened 2021-03-01
Drop usage of outdated protobuf module

markusthoemmes:
I love me a good cleanup and github.com/golang/protobuf says

> It has been superseded by the google.golang.org/protobuf module, which contains an updated and simplified API, support for protobuf reflection, and many other improvements. We recommend that new code use the google.golang.org/protobuf module.

So, why not? It feels cleaner too and the wrapper helpers are nice :).


davidor (approved):
/lgtm

PR #476, opened 2021-03-22
Replace copied ingress CRD with link to networking one

markusthoemmes:
# Changes

Instead of having a hard copy of the Ingress CRD inside the repo, this links to the respective CRD in the vendor dir to avoid duplication and divergence.

/assign @nak3 @jmprusi @davidor


davidor (approved):
/lgtm

PR #486, opened 2021-03-31
Globally reconcile conflicted Ingresses on deletion

markusthoemmes:
# Changes

Tested in https://github.com/knative/serving/pull/11091

If an Ingress goes into a conflicting state it might never come back from that because nothing triggers its reconcilation again. Most notably: The deletion of the Ingress that caused the conflict should trigger reconcilation.

So this adds a global reconcilation of all Ingresses that are in a conflict state when any Ingress is removed.

**Release Note**

```release-note
Fixed a bug where an Ingress that is in a conflict never resumes from that state without kicking it.
```


nak3:
/lgtm
Thank you!

nak3 (approved):

PR #497, opened 2021-04-20
Fix potentially racy ingress deletion behavior

### LINKED ISSUE #448: Finalization logic seems incorrect
I'm looking at this logic here:
https://github.com/knative-sandbox/net-kourier/blob/48320f242b9a2a8fd09d4c5ca29a0f0bfd447046/pkg/reconciler/ingress/ingress.go#L95-L110

The use of finalizers here seems both overkill and buggy, I'll try to elaborate on both facets of this below.

The use of finalizers here is overkill because we're just trying to make sure that the *in-memory* index is updated to note the removal of the kingress.  Finalizers are generally needed when state outside of the controller process needs cleaning up, and that's not the case here.

The use of finalizers here is buggy because only the "main" controller is guaranteed to process the deletion (since it removes the finalizer itself).

My recommendation here would be that since both the informer cache and the envoy configuration index are both in-memory, that it is both necessary and sufficient to directly delete the configuration in the `DeleteFunc` informer event and drop the use of finalizers here.  We (@vaikas) made a similar change to the in-memory channel in eventing relatively recently for similar reasons.

markusthoemmes:
# Changes

As per title really. The linked issue has all the context as to why this is necessary.

/kind <kind>

Fixes #448

/hold for the change in networking

### DIFF
--- a/pkg/reconciler/ingress/ingress.go
+++ b/pkg/reconciler/ingress/ingress.go
@@ -21,6 +21,7 @@ import (
 	"errors"
 	"fmt"
 
+	"k8s.io/apimachinery/pkg/types"
 	"knative.dev/net-kourier/pkg/config"
 	envoy "knative.dev/net-kourier/pkg/envoy/server"
 	"knative.dev/net-kourier/pkg/generator"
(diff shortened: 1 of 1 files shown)


pkg/reconciler/ingress/ingress.go
```diff
@@ -108,23 +109,24 @@ func (r *Reconciler) ObserveKind(ctx context.Context, ing *v1alpha1.Ingress) rec
...
-	if err := r.caches.DeleteIngressInfo(ctx, ing.Name, ing.Namespace); err != nil {
+	if err := r.caches.DeleteIngressInfo(ctx, key.Name, key.Namespace); err != nil {
 		return err
 	}
 
 	return r.updateEnvoyConfig(ctx)
 }
 
+func (r *Reconciler) FinalizeKind(ctx context.Context, ing *v1alpha1.Ingress) reconciler.Event {
+	// Keeping this for now to keep finalizer based logic intact.
```
julz, line 126: I don't quite understand how keeping this method helps here, can you say a couple words more?

markusthoemmes: We won't actively remove the finalizer, so if we drop this function I believe the object will get stuck in finalization forever.

julz: Aah, I see, what a pain - yeah makes sense

nak3 (approved):
/lgtm

PR #502, opened 2021-04-26
Move conflict-resync into ObserveDeletion handler

markusthoemmes:
# Changes

Without the resync being triggered by the ObserveDeletion handler, we
might have a race between triggering the resync and actually reconciling
the deletion of an Ingress. So the resynced keys might see a conflict
yet again because the deletion was not yet reconciled when the resync
happened. This puts things into the correct order.

/hold

To try this out properly in Serving.


markusthoemmes:
Passed consistently on https://github.com/knative/serving/pull/11261.

/unhold

/assign @jmprusi @davidor @nak3

jmprusi (approved):

PR #516, opened 2021-05-17
Replace golint with revive

markusthoemmes:
# Changes

See https://groups.google.com/g/golang-nuts/c/rCP70Aq_tBc/m/8QHp6_cqBgAJ
as towards golint's deprecation. revive is supposed to be a drop-in
replacement.


davidor (approved):
/lgtm

PR #521, opened 2021-05-20
Create basic config map machinery

markusthoemmes:
# Changes

Ref #518 

This adds the first steps to support configuring Kourier through a ConfigMap with the first goal being able to enable and disable access logging respectively.

PR #523, opened 2021-05-20
Create config-kourier and hook it all up

### LINKED ISSUE #518: Access logging should be configurable
Currently, we log every single request going through Kourier. For anything not trivial, that can quickly cause issues with the logging volume. We should at least make this behavior configurable if not even make the default to not log.

markusthoemmes:
# Changes

Fixes #518

This adds a new ConfigMap `config-kourier`, with the initial `enable-service-access-logging` setting, which allows control over the access log behavior of Kourier at runtime.

### DIFF
--- a/config/200-config.yaml
+++ b/config/200-config.yaml
@@ -38,5 +38,6 @@ data:
     # to actually change the configuration.
 
     # Specifies whether requests reaching the Kourier gateway
-    # should be logged.
-    enable-access-logging: true
+    # in the context of services should be logged. Readiness
+    # probes etc. must be configured via the bootstrap config.
+    enable-service-access-logging: "true"
--- a/pkg/config/configmap.go
+++ b/pkg/config/configmap.go
@@ -26,13 +26,14 @@ const (
 	// ConfigName is the name of config map for Kourier.
 	ConfigName = "config-kourier"
 
-	// enableAccessLoggingKey is the config map key for queue sidecar image.
-	enableAccessLoggingKey = "enable-access-logging"
+	// enableServiceAccessLoggingKey is the config map key for enabling service related
+	// access logging.
+	enableServiceAccessLoggingKey = "enable-service-access-logging"
 )
 
 func DefaultConfig() *Kourier {
 	return &Kourier{
-		EnableAccessLogging: true, // true is the default for backwards-compat
+		EnableServiceAccessLogging: true, // true is the default for backwards-compat
 	}
 }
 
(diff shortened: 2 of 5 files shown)


markusthoemmes:
This contains #521. We can merge them stepwise or straight this one, as you'd prefer in review :)

config/200-config.yaml
```diff
@@ -0,0 +1,42 @@
...
+    # options and document them in a way that is accessible
+    # to users that `kubectl edit` this config map.
+    #
+    # These sample configuration options may be copied out of
+    # this example block and unindented to be in the data block
+    # to actually change the configuration.
+
+    # Specifies whether requests reaching the Kourier gateway
+    # should be logged.
+    enable-access-logging: true
```
nak3, line 42:
```suggestion
    enable-access-logging: "true"
```

Without `""` quote, the configuration fails to be recognized.

nak3 (commented):
LGTM except for one minor issue.

nak3:
One note I noticed during test. (Not an issue and I guess we cannot fix it.)

Even when we set `enable-access-logging` to false, Probe's access log does not stop:

```
[2021-05-21T09:07:20.850Z] "GET /ready HTTP/1.1" 200 - 0 5 0 - "-" "kube-probe/1.20" "3cee842a-79c6-4c77-83c1-4751011137d4" "internalkourier" "-"
[2021-05-21T09:07:20.850Z] "GET /ready HTTP/1.1" 200 - 0 5 0 0 "-" "kube-probe/1.20" "3cee842a-79c6-4c77-83c1-4751011137d4" "internalkourier" "/tmp/envoy.admin"
  ... continue ...
```

this is because readiness probe uses unix domain socket for the access and envoy's log setting does not work for it.

https://github.com/knative-sandbox/net-kourier/blob/40378523de25a3ff6ec067b29c4a7149f9f89aad/config/300-gateway.yaml#L64

markusthoemmes:
That's right @nak3, I actually wanted to send a separate PR that removes the logging on the admin endpoint completely (it's configured in the bootstrap config, so a bit hard to connect to this setting I think). Should I rename the setting to `enable-service-access-logging` maybe, to clarify the difference?

nak3:
I like both :smile: `enable-service-access-logging` is clear but `enable-access-logging` is also short and fine for service access log.

markusthoemmes:
@nak3 changed the keys for clarity!

nak3 (approved):
/lgtm

PR #555, opened 2021-06-18
rename various things to be consistent with net-* components

dprotaso:
Part of https://github.com/knative/networking/issues/448

### DIFF
--- a/config/200-serviceaccount.yaml
+++ b/config/200-serviceaccount.yaml
@@ -15,7 +15,7 @@
 apiVersion: v1
 kind: ServiceAccount
 metadata:
-  name: 3scale-kourier
+  name: net-kourier
   namespace: knative-serving
   labels:
     networking.knative.dev/ingress-provider: kourier
@@ -24,7 +24,7 @@ metadata:
 apiVersion: rbac.authorization.k8s.io/v1
 kind: ClusterRole
 metadata:
-  name: 3scale-kourier
+  name: net-kourier
   namespace: knative-serving
   labels:
     networking.knative.dev/ingress-provider: kourier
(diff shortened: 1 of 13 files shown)


config/200-serviceaccount.yaml
```diff
@@ -15,7 +15,7 @@
 apiVersion: v1
 kind: ServiceAccount
 metadata:
-  name: 3scale-kourier
+  name: net-kourier
```
dprotaso, line 18: Curious any reason why you don't opt into the aggregated cluster role?

config/300-controller.yaml
```diff
@@ -24,15 +24,15 @@ spec:
...
   template:
     metadata:
       labels:
-        app: 3scale-kourier-control
+        app: net-kourier-controller
     spec:
       containers:
         - image: ko://knative.dev/net-kourier/cmd/kourier
-          name: kourier-control
+          name: controller
```
nak3, line 35: I'm fine for this change, but users have to change `spec.registry.override` if they configured :sweat_drops: . In addition, they need to change it `deployment/container` style like  https://github.com/knative/docs/pull/3821/files

config/200-serviceaccount.yaml
nak3: I think no reason. We can use aggregated cluster role.

nak3:
LGTM

dprotaso:
So the gateway deployment is renamed but the labels will remain the same

dprotaso:
@nak3 @markusthoemmes this is ready now

pkg/reconciler/ingress/controller.go
```diff
@@ -51,9 +51,9 @@ import (
 
 const (
 	gatewayLabelKey   = "app"
-	gatewayLabelValue = "3scale-kourier-gateway"
+	gatewayLabelValue = "net-kourier-gateway"
```
markusthoemmes, line 54: This is wrong I believe, as you haven't changed the label for the gateways.

markusthoemmes (commented):
Verified this works as intended locally. Verified that the config change does not impact the running cluster.

markusthoemmes:
Gonna hold this for this release. Manual application has been verified but I'd really like the operator to behave nicely with the changes as well. Currently, that's not been verified and I'm quite sure it breaks in that traffic will be dropped. I'll attend to this PR more closely to target the next release for landing. Sorry for not getting here earlier.

/hold

markusthoemmes (approved):
/lgtm
/approve
/unhold

PR #570, opened 2021-07-12
Implement a health check for the controller

### LINKED ISSUE #560: Add a readiness probe to kourier-control
When rolling the deployments (i.e. in an upgrade), the deployment rollout cannot tell if the new control pods are **really** up yet. We should add a readiness probe that keys off of the gRPC server being serving correctly.

/assign

markusthoemmes:
Fixes #560

As per title, this adds a health check to the controller so that it only becomes ready once the gRPC server actually started serving. That may be important in upgrade cases where Kubernetes currently basically rolls the controllers "immediately" even if the new ones aren't properly up yet, resulting in an intermittent state of no control-plane availability.


nak3:
LGTM except for the boilerplate error.

nak3 (approved):
/lgtm

PR #572, opened 2021-07-13
Bump a few assorted deps

markusthoemmes:
As per title. go-control-plane is sadly a "little" wedged as they dropped support for v2. We shall look at that at some point.


nak3 (approved):
/lgtm

PR #574, opened 2021-07-14
Wait for rollout of new controllers after updating env

markusthoemmes:
The test changes the environmenrt variables of the controller deployment but then doesn't wait until the new values are rolled out. This got made worse by us adding a readiness probe which prolongs the rollout.

This should make things consistent again.


nak3 (approved):
/lgtm

PR #584, opened 2021-07-29
Do not add target endpoints twice for probe

nak3:
This patch removes stops adding target endpoints twice.

Currently the target endpoint for probe is added twice as:

https://github.com/knative-sandbox/net-kourier/blob/d3a2c5b5a60e32baea0b6c92c6e78430fe24e0c2/pkg/reconciler/ingress/lister.go#L84

https://github.com/knative-sandbox/net-kourier/blob/d3a2c5b5a60e32baea0b6c92c6e78430fe24e0c2/pkg/reconciler/ingress/lister.go#L93

It should be enough to add only once.


markusthoemmes (approved):
How on earth did you discover that? :joy: 

/approve
/lgtm

PR #587, opened 2021-08-02
Remove unused script to setup kind cluster

markusthoemmes:
The fact that Knative Serving is 6 releases old in this script and nobody cared tells me nobody is actually using this.


nak3 (approved):
/lgtm

PR #434, opened 2021-02-02
WIP - Use envoy v3 API 

jmprusi:
# Changes

- Uses Envoy v3 API instead of the deprecated v2.

/kind enhancement

**Release Note**

```release-note
Use envoy v3 API instead of the deprecated v2 version. 
```

PR #598, opened 2021-08-18
Use envoy v3 API

nak3:
This patch takes over https://github.com/knative-sandbox/net-kourier/pull/434

/kind enhancement

**Release Note**

```release-note
Use envoy v3 API instead of the deprecated v2 version. 
```


markusthoemmes:
It's awesome that this passes :tada:! Also awesome that the upgrade tests are passing. Have you done some manual testing by any chance as well? I'd especially be curious what happens if a gateway with the old config tries to talk to a new controller. Does that work?

nak3:
Oh, that's good idea. I haven't tried with the old config. Let me try to test it. (Probably tomorrow.)

nak3:
I confirmed. Gateway with the old config does not work as `transport_api_version` and `resource_api_version` needs to set `V3` explicitly. There is no way to set both api versions v2 and v3 as per [doc](https://www.envoyproxy.io/docs/envoy/latest/api-v3/config/core/v3/config_source.proto#enum-config-core-v3-apiversion).

So, when gateway has the old (incompatible) version config, controller cannot set the route path [`/ready`](https://github.com/knative-sandbox/net-kourier/blob/a20e6f46b5d707076c81845fc0a6595e50a9491f/pkg/generator/status_vhost.go#L57) for the readiness probe and gateway cannot become Ready as:

```
$ kubectl -n kourier-system  get pod -w
NAME                                      READY   STATUS    RESTARTS   AGE
3scale-kourier-gateway-6d8f6b8549-95xgn   0/1     Running   0          3m22s

$ kubectl -n kourier-system  describe pod
  ...
  Warning  Unhealthy  71s (x22 over 2m56s)  kubelet            Readiness probe failed: Get "http://10.244.1.51:8081/ready": dial tcp 10.244.1.51:8081: connect: connection refused
```

markusthoemmes:
Ah okay, I see. So when we do an upgrade and change the configMap immediately, do you know how that's working correctly then? It might be that envoy loses connection to the controller and can't connect momentarily but it keeps its existing configuration until it can successfully reconnect, which it eventually can. If that's the case, I think this should be safe.

nak3:
I found a coulple of things:
- the ConfigMap (`envoy-bootstrap.yaml`) is a static config and does not affect until gateway pod was restarted.
- 300-gateway.yaml has no change from previous upgrade so gateway pod is not restarted during upgrade test.

So, the ingress created before the upgrade works fine with old gateway pod during upgrade. But I guess a new ingress will not work as controller cannot connect to the gateway after the upgrade. I am confirming by adding the test.

nak3:
Upgrade test passed because gateway has no change during the upgrade. But it fails to create a new ingress.
Restarting the Gateway pod manually fixes the problem.
So, I think we just need to write a release note taht users need to restart gateway pod by themselves.

### 1. Before update

```
+------------+           +---------+   +-----------+
| controller | --- o --> | gateway | - | bootstrap |
| api v2     |           |         |   | api v2    |
+------------+		 +---------+   +-----------+
```

### 2. Apply update

* Gateway still can handle client requests.
* But controller cannot connect to gateway so users cannot create/update a ksvc(kingress).

```
+------------+		 +---------+   +-----------+
| controller | --- x --> | gateway | - | bootstrap |
| api v3     |           |         |   | api v2    |
+------------+		 +---------+   +-----------+

                                       +-----------+
               	                       | bootstrap |
               			       | v3	   |
               			       +-----------+
                                 (the new config is not used)
```

### 3. Restart gateway (manually)

* gateway starts using the new config and accept the connection from controller.

```
+------------+		 +---------+   +-----------+
| controller | --- o --> | gateway | - | bootstrap |
| api v3     |           |         |   | api v3    |
+------------+		 +---------+   +-----------+
```

markusthoemmes:
Awesome, thanks for the analysis! Can we as an alternative also add annotation or whatever to the Gateway pod template to force a rollout? I.e. add a `networking.knative.dev/poke: "0.26"` to signify that the pod **has** to be restarted? Alternativelty, I could also imagine a mechanism like `networking.knative.dev/bootstrapHash: abcdf` and fill in the hashed bootstrap configMap to have a generalized way of forcing a bootstrap config update.

I'd personally be fine with the more simple approach for now too tho.

nak3:
`networking.knative.dev/bootstrapHash: abcdf` will be complicated so I added `networking.knative.dev/poke: "v0.26"` to the gateway pod to restart.
The drawback is that gateway pod will be restarted when we drop the label on v0.27 :sweat:

markusthoemmes:
We don't have to drop the label :) (let's make it an annotation btw, we don't need to filter by this I think). We can keep it to signify: "This should be poked because of a change in 0.26" so even if you bump from 0.25 to 0.27, that'd still be valid. We can update it, once we need another poke or thelike, no biggy imo.

Issue #428, opened 2021-01-27
support for httpProtocol: "Redirected"


mjgallag:
[httpProtocol: "Redirected"](https://github.com/knative/networking/blob/94433ab7f6202fa21049dde4afa5af6cfb25b705/config/config-network.yaml#L106)

In the interim, a workaround to enable this directly in kourier would be greatly appreciated, thanks!

nak3:
The option should be supported by https://github.com/knative-sandbox/net-kourier/pull/619.

PR #619, opened 2021-08-29
Support `httpProtocol: Redirected`

nak3:
This patch supports `Spec.HTTPOption` in KIngress as well as
`httpProtocol: Redirected` in config-network.

Note about internal stuff as the redirect support needs some
internal changes:

Currently RouteConfig (`external_services`) is used by both HTTP
Listener and HTTPS Listener when TLS is enabled.
However, to add redirect support, we cannot share the RouteConfig as
HTTPListener needs Route with `RedirectAction`.

So, this patch introduces another RouteConfig
(`external_tls_services`) for the HTTPS Listener.

Fix https://github.com/knative/networking/issues/304 https://github.com/knative-sandbox/net-kourier/issues/428

**Release Note*

```release-note
net-kourier supports `httpProtocol: Redirected` in config-network.
```

### DIFF
--- a/pkg/generator/ingress_translator.go
+++ b/pkg/generator/ingress_translator.go
@@ -175,10 +175,16 @@ func (translator *IngressTranslator) translateIngress(ctx context.Context, ingre
 			}
 
 			if len(wrs) != 0 {
-				routes = append(routes, envoy.NewRoute(
-					pathName, matchHeadersFromHTTPPath(httpPath), path, wrs, 0, httpPath.AppendHeaders, httpPath.RewriteHost))
+				if ingress.Spec.HTTPOption == v1alpha1.HTTPOptionRedirected && rule.Visibility == v1alpha1.IngressVisibilityExternalIP {
+					routes = append(routes, envoy.NewRedirectRoute(
+						pathName, matchHeadersFromHTTPPath(httpPath), path))
+				} else {
+					routes = append(routes, envoy.NewRoute(
+						pathName, matchHeadersFromHTTPPath(httpPath), path, wrs, 0, httpPath.AppendHeaders, httpPath.RewriteHost))
+				}
 				if len(ingress.Spec.TLS) != 0 {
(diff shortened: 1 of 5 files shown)


markusthoemmes:
I've tried to review this 3 times now but struggle to gauge the impact :cry:. Do you think we can split this into 2 PRs, where the first one would just do the split between HTTP and HTTPS stuff and the second adds the redirect route? Or is that not possible?

nak3:
Sure, no problem. I sent the part 1 (split between HTTP and HTTPS stuff ) https://github.com/knative-sandbox/net-kourier/pull/626

pkg/generator/ingress_translator.go
```diff
@@ -175,10 +175,16 @@ func (translator *IngressTranslator) translateIngress(ctx context.Context, ingre
...
-					pathName, matchHeadersFromHTTPPath(httpPath), path, wrs, 0, httpPath.AppendHeaders, httpPath.RewriteHost))
+				if ingress.Spec.HTTPOption == v1alpha1.HTTPOptionRedirected && rule.Visibility == v1alpha1.IngressVisibilityExternalIP {
+					routes = append(routes, envoy.NewRedirectRoute(
+						pathName, matchHeadersFromHTTPPath(httpPath), path))
+				} else {
+					routes = append(routes, envoy.NewRoute(
+						pathName, matchHeadersFromHTTPPath(httpPath), path, wrs, 0, httpPath.AppendHeaders, httpPath.RewriteHost))
+				}
 				if len(ingress.Spec.TLS) != 0 {
-					tlsRoutes = append(tlsRoutes, routes...)
```
markusthoemmes, line 181: This makes me wonder: Was this correct before? :thinking: :joy:

nak3: Before this PR, `tlsRoutes` and `routes` have same values so I think it was Okay :thinking:

markusthoemmes (approved):
/lgtm
/approve

PR #626, opened 2021-09-03
Split RouteConfig for HTTP and HTTPS listeners

nak3:
Currently RouteConfig (`external_services`) is used by both HTTP
Listener and HTTPS Listener when TLS is enabled.
However, to add redirect support, we cannot share the RouteConfig as
HTTPListener needs Route with `RedirectAction`.

So, this patch introduces another RouteConfig
(`external_tls_services`) for the HTTPS Listener.

Part of https://github.com/knative-sandbox/net-kourier/pull/619

### DIFF
--- a/pkg/generator/caches.go
+++ b/pkg/generator/caches.go
@@ -39,12 +39,13 @@ import (
 )
 
 const (
-	envCertsSecretNamespace = "CERTS_SECRET_NAMESPACE"
-	envCertsSecretName      = "CERTS_SECRET_NAME"
-	certFieldInSecret       = "tls.crt"
-	keyFieldInSecret        = "tls.key"
-	externalRouteConfigName = "external_services"
-	internalRouteConfigName = "internal_services"
+	envCertsSecretNamespace    = "CERTS_SECRET_NAMESPACE"
+	envCertsSecretName         = "CERTS_SECRET_NAME"
+	certFieldInSecret          = "tls.crt"
+	keyFieldInSecret           = "tls.key"
+	externalRouteConfigName    = "external_services"
+	externalTLSRouteConfigName = "external_tls_services"
+	internalRouteConfigName    = "internal_services"
 )
 
 // ErrDomainConflict is an error produces when two ingresses have conflicting domains.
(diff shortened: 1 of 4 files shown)


pkg/generator/ingress_translator.go
```diff
@@ -174,6 +177,8 @@ func (translator *IngressTranslator) translateIngress(ctx context.Context, ingre
 			if len(wrs) != 0 {
 				routes = append(routes, envoy.NewRoute(
 					pathName, matchHeadersFromHTTPPath(httpPath), path, wrs, 0, httpPath.AppendHeaders, httpPath.RewriteHost))
+				tlsRoutes = append(tlsRoutes, envoy.NewRoute(
```
markusthoemmes, line 180: Do we really want to always do this, even if the Ingress has no TLS config?

pkg/generator/ingress_translator.go
```diff
@@ -182,20 +185,34 @@ func (translator *IngressTranslator) translateIngress(ctx context.Context, ingre
 			return nil, nil
 		}
 
-		var virtualHost *route.VirtualHost
+		if len(ingress.Spec.TLS) != 0 {
+			tlsRoutes = append(tlsRoutes, routes...)
```
nak3, line 189: This line (Copy `routes` to `tlsRoutes`) will not work as https://github.com/knative-sandbox/net-kourier/pull/619/  needs the condition - routes and tlsRoutes are not same when Redirect option.

pkg/generator/ingress_translator.go
nak3: We don't need it. Updated the PR to create tlsRoutes only when ingress has TLS config.
There are some extra `if` conditions so I always created tlsRoutes in the first PR.

pkg/generator/caches.go
```diff
@@ -238,21 +246,24 @@ func generateListenersAndRouteConfigs(
...
+			ctx, externalTLSManager, kubeclient,
 		)
 		if err != nil {
 			return nil, nil, err
 		}
 		listeners = append(listeners, externalHTTPSEnvoyListener)
 	}
 
+	if len(sniMatches) > 0 || useHTTPSListenerWithOneCert() {
+		return listeners, []cachetypes.Resource{externalRouteConfig, externalTLSRouteConfig, internalRouteConfig}, nil
```
markusthoemmes, line 265: nit: Should we create a `resources` list which we append to in the respective branches so we avoid this branch here? envisioning this to be `return listeners, resources, nil` (and without the `if`)

markusthoemmes (approved):
/lgtm
/approve

/hold
For the actions to pass.

nak3:
/hold cancel

All tests passed :tada:

PR #634, opened 2021-09-09
Set ValidateClusters in NewRouteConfig

nak3:
Current code adds the validation one by one, but we always need it.
So this patch sets ValidateClusters in NewRouteConfig by default.

/kind cleanup


markusthoemmes (approved):
/lgtm
/approve

PR #650, opened 2021-09-29
Update dependencies (mostly go-control-plane)

markusthoemmes:
# Changes

As per title. Bumping gRPC deps bumps go-control-plane and that brings a few API changes with it.


nak3 (approved):
/lgtm

PR #667, opened 2021-10-18
create tls route when using one cert via environment value.

### LINKED ISSUE #665: TLS setup is not working in release 0.26?
Hello,
I think there is some kind of regression with TLS setup, I deployed release 0.26 with a knative 0.24, and when I curl my function I got `404 http code`.

Looking at the code, the route for TLS is created only if `len(ingress.Spec.TLS) != 0 `, but I think `ingressesToSync, err := getReadyIngresses(ctx, knativeClient.NetworkingV1alpha1())` is returning an ingress without any TLS specifications.

As far as I understand, we don't have to match SNI when configuring TLS with a single cert for all services, so patching the code as following in `ingress_translator.go`, has created the route
before:
```
if len(ingress.Spec.TLS) != 0 {
	tlsRoutes = append(tlsRoutes, envoy.NewRoute(
		pathName, matchHeadersFromHTTPPath(httpPath), path, wrs, 0, httpPath.AppendHeaders, httpPath.RewriteHost))
}
```
after:
```
if len(ingress.Spec.TLS) != 0 || useHTTPSListenerWithOneCert() {
	tlsRoutes = append(tlsRoutes, envoy.NewRoute(
		pathName, matchHeadersFromHTTPPath(httpPath), path, wrs, 0, httpPath.AppendHeaders, httpPath.RewriteHost))
}
```
Is this a real issue or I'm doing something wrong? or maybe because I 'm using knative 0.24?

Of course, I've create a secret with a cert and key and added the required env variables

Thanks

RamyChaabane:
# Changes

Create tls route when using one cert via environment value.

/kind bug

Fixes #665

### DIFF
--- a/pkg/generator/ingress_translator.go
+++ b/pkg/generator/ingress_translator.go
@@ -182,7 +182,7 @@ func (translator *IngressTranslator) translateIngress(ctx context.Context, ingre
 					routes = append(routes, envoy.NewRoute(
 						pathName, matchHeadersFromHTTPPath(httpPath), path, wrs, 0, httpPath.AppendHeaders, httpPath.RewriteHost))
 				}
-				if len(ingress.Spec.TLS) != 0 {
+				if len(ingress.Spec.TLS) != 0 || useHTTPSListenerWithOneCert() {
 					tlsRoutes = append(tlsRoutes, envoy.NewRoute(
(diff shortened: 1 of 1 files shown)


nak3:
/ok-to-test

@RamyChaabane Thank you for the fix. I verified the fix in my local and we will add the e2e test https://github.com/knative-sandbox/net-kourier/pull/669 once this PR was merged.

Could you please try to comment `@googlebot I fixed it.`? It seems CLA still says "no".

PR #669, opened 2021-10-19
Add e2e test for one certs via env value

nak3:
Currently there are no e2e test for one certs via env value so we may contain some breaking change.
This patch adds the e2e test.


test/e2e/certificate_test.go
```diff
@@ -0,0 +1,147 @@
...
+	// Check without TLS.
+	ingress.RuntimeRequest(ctx, t, client, "http://"+name+".example.com")
+
+	// Check with TLS.
+	ingress.RuntimeRequest(ctx, t, client, "https://"+name+".example.com")
+
+}
+
+// updateDeploymentEnv updates Deployment's env values.
+func updateDeploymentEnv(ctx context.Context, clients *test.Clients, certsNamespace, certsName string) error {
```
markusthoemmes, line 115: Should we pull this into the script? We do that with ExtAuth stuff already.

PR #694, opened 2021-11-05
Adds k8s recommended labels

psschwei:
# Changes

Adds Kubernetes recommended labels to networking as discussed in https://github.com/knative/networking/issues/552

Note that changes to the labels in `vendor/knative.dev/networking` should be handled by the automation once https://github.com/knative/networking/pull/555 merges. 

/kind enhancement

/assign @dprotaso

### DIFF
--- a/config/100-namespace.yaml
+++ b/config/100-namespace.yaml
@@ -18,7 +18,7 @@ metadata:
   name: kourier-system
   labels:
     networking.knative.dev/ingress-provider: kourier
-    app.kubernetes.io/name: kourier
+    app.kubernetes.io/name: knative-serving
+    app.kubernetes.io/component: net-kourier
     app.kubernetes.io/version: devel
-    app.kubernetes.io/part-of: knative-serving
     serving.knative.dev/release: devel
--- a/config/100-namespace.yaml
+++ b/config/100-namespace.yaml
@@ -18,6 +18,7 @@ metadata:
   name: kourier-system
   labels:
     networking.knative.dev/ingress-provider: kourier
+    app.kubernetes.io/name: kourier
     app.kubernetes.io/version: devel
     app.kubernetes.io/part-of: knative-serving
     serving.knative.dev/release: devel
--- a/config/100-namespace.yaml
+++ b/config/100-namespace.yaml
@@ -18,4 +18,6 @@ metadata:
   name: kourier-system
   labels:
     networking.knative.dev/ingress-provider: kourier
+    app.kubernetes.io/version: devel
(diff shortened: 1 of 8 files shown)


hack/release.sh
```diff
@@ -35,8 +35,8 @@ readonly RELEASES
 function build_release() {
   # Update release labels if this is a tagged release
   if [[ -n "${TAG}" ]]; then
-    echo "Tagged release, updating release labels to serving.knative.dev/release: \"${TAG}\""
-    LABEL_YAML_CMD=(sed -e "s|serving.knative.dev/release: devel|serving.knative.dev/release: \"${TAG}\"|")
+    echo "Tagged release, updating release labels to serving.knative.dev/release: \"${TAG}\" and app.kubernetes.io/version: \"${TAG}\""
+    LABEL_YAML_CMD=(sed -e "s|serving.knative.dev/release: devel|serving.knative.dev/release: \"${TAG}\"|" -e "s|app.kubernetes.io/version: devel|app.kubernetes.io/version: \"${TAG:1}\"|")
```
nak3, line 39: @psschwei @dprotaso 
It is not possible to create `build_release()` or `add_label()` function in  https://github.com/knative/hack/blob/main/release.sh and share it with all net-* repos?

dprotaso: yeah - but I'd do that in a separate PR

test/config/chaosduck.yaml
```diff
@@ -24,6 +24,9 @@ metadata:
   name: chaosduck
   namespace: kourier-system
   labels:
+    app.kubernetes.io/name: chaosduck
+    app.kubernetes.io/version: devel
+    app.kubernetes.io/part-of: knative-serving
```
dprotaso, line 29: Any reason why you added it to the `test/config` resources? We don't release these.

dprotaso (approved):

config/100-namespace.yaml
```diff
@@ -18,4 +18,7 @@ metadata:
   name: kourier-system
   labels:
     networking.knative.dev/ingress-provider: kourier
+    app.kubernetes.io/name: kourier
```
dprotaso, line 21: unsure if you want to be consistent with the other repos and call this `net-kourier`

nak3: +1 for `net-kourier`.

test/config/chaosduck.yaml
psschwei: Only because they showed up in my rgrep... will remove it

PR #704, opened 2021-11-15
Introduce `KOURIER_HTTPOPTION_DISABLED` env variable to disable redirect

nak3:
This patch introduces `KOURIER_HTTPOPTION_DISABLED` env variable to stop creating redirect route when it is set.
This option is useful when front end proxy handles the redirection by `spec.HTTPOption`.
e.g. Kourier on OpenShift handles `HTTPOption` by OpenShift Route so `KOURIER_HTTPOPTION_DISABLED` should be set.

### DIFF
--- a/pkg/generator/ingress_translator_test.go
+++ b/pkg/generator/ingress_translator_test.go
@@ -17,7 +17,6 @@ limitations under the License.
 package generator
 
 import (
-	"os"
 	"testing"
 	"time"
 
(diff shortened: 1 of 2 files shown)


nak3:
/cc @markusthoemmes 

Could you please take a look? And do you have any better idea for https://issues.redhat.com/browse/SRVKS-840 ?

pkg/generator/ingress_translator_test.go
```diff
@@ -546,6 +547,180 @@ func TestIngressTranslator(t *testing.T) {
...
+				externalVirtualHosts:    []*route.VirtualHost{},
+				externalTLSVirtualHosts: []*route.VirtualHost{},
+				internalVirtualHosts:    vHosts,
+			}
+		}(),
+	}}
+
+	for _, test := range tests {
+		t.Run(test.name, func(t *testing.T) {
+			os.Setenv("KOURIER_HTTPOPTION_DISABLED", "true")
```
markusthoemmes, line 696: Since we switched to Golang 1.17, we can use `t.Setenv` I believe, which does the unsetting for us?

markusthoemmes (commented):
LGTM modulo the test nit.

pkg/generator/ingress_translator_test.go
nak3: Yes, it works. Thank you! Updated.

markusthoemmes (approved):
/lgtm
/approve

PR #714, opened 2021-11-20
Clean up test/e2e-common.sh

nak3:
This patch cleans up test/e2e-common.sh.

Current script does:
- scale down to `0` then scale up to `2`.
- wait for leader controller.

But the same test in `test/e2e-kind.sh` is working without such
operation.

So this patch removes them.


markusthoemmes (approved):
/lgtm
/approve

PR #730, opened 2021-12-07
[WIP][POC] Add multiple Kourier support

nak3:
This patch supports multiple Kourier support.

The idea is that adding `KOURIER_MEMBER_ROLL` env value into controller to filter the
namespaces where Ingresses are reconciled.

```
            - name: KOURIER_MEMBER_ROLL
              value: "test-ns1,test-ns2"
```

The code change is just a few lines as https://github.com/knative-sandbox/net-kourier/pull/730/commits/eac1fcf511d42dc7fb719629523d39e81c60afa0 and no change current behavior
when `KOURIER_MEMBER_ROLL` was not defined in the yaml.

This PR contains a simple e2e test with one more Kourier manifests so it is a big.

### DIFF
--- a/pkg/reconciler/ingress/controller.go
+++ b/pkg/reconciler/ingress/controller.go
@@ -18,6 +18,7 @@ package ingress
 
 import (
 	"context"
+	"os"
 	"strings"
 
 	v3 "github.com/envoyproxy/go-control-plane/envoy/service/discovery/v3"
@@ -63,6 +64,8 @@ var isKourierIngress = reconciler.AnnotationFilterFunc(
 	v1alpha1ingress.ClassAnnotationKey, config.KourierIngressClassName, false,
 )
 
+var kourierChainFilter func(interface{}) bool
+
 func NewController(ctx context.Context, cmw configmap.Watcher) *controller.Impl {
 	logger := logging.FromContext(ctx)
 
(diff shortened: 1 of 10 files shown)


pkg/reconciler/ingress/controller.go
```diff
@@ -74,6 +77,14 @@ func NewController(ctx context.Context, cmw configmap.Watcher) *controller.Impl
 	podInformer := podinformer.Get(ctx)
 	secretInformer := secretinformer.Get(ctx)
 
+	// Add namespace filter to be reconciled by this controller.
+	namespaceFilters := []func(interface{}) bool{}
+	for _, namespace := range strings.Split(os.Getenv("KOURIER_MEMBER_ROLL"), ",") {
```
matzew, line 82: could this done more dynamically with like labeling of namespaces?

nak3: I think it is possible, but labeling of namespaces will not be "secure" multi-tenant, will it? It is possible to hijack any tenants by adding the label to a namespace of their own.

skonto: Istio is also trying to support multi-tenancy: https://github.com/istio/api/pull/2212. Label selectors do provide an option for filtering for performance reasons (https://istio.io/latest/blog/2021/discovery-selectors) but need mesh membership verification to be valid if used. Performance means also filter objects at the informer level this PR adds filtering for the reconciler work.

Labels can be used for labeling tenants check [here](https://www.tigera.io/blog/label-standard-and-best-practices-for-kubernetes-security/) and are useful for grouping objects. It is also the case that you need to protect certain labels from being applied if you go down that path with labels like with `kubernetes.*` labels.

I think we need to implement a model where:

a) security is in place for tenants under some multi-tenancy model (istio has a number of deployment models)
b) optimize K8s object listing and reconciliation per net-kourier instance

For b) the issue of how to dynamically update the list can be solved either with:
a) updating the deployment of the net-kourier wrt to some configmap where the ns list resides. Eg could work with the Knative operator.
b) Use an an informer factory to create dynamically informers linked to the controller with a function similar to [this](https://github.com/knative/pkg/blob/f907b31046d3c69a29b13d6d6ad2907767faac1e/configmap/informer/informed_watcher.go#L52).That would require to handle potentially the informer lifecycle and could cause issues with cache re-population (can take time).

PR #743, opened 2022-01-07
Fix TLS listener filter chains when configured with both a secret (via env vars) and SNI

### LINKED ISSUE #740: Global TLS is disabled when an KIngress is configured with a TLS secret
Serving Issue: https://github.com/knative/serving/issues/12377

Global TLS is configured here:
https://github.com/knative-sandbox/net-kourier#setup-tls-certificate

When you create a KIngress with it's own TLS secret (ie. provided by autoTLS) it elides the configured global tls cert from the envoy's filter chain.

Looks like it's elided here when the SNI listener is created
https://github.com/knative-sandbox/net-kourier/blob/87c1085d407b1a5026cd231eba64c90ca91ec7f4/pkg/generator/caches.go#L237-L256

norbjd:
# Changes

When TLS is configured with 1 or more SNI, and if a certificate is set via environment variables (`CERTS_SECRET_NAMESPACE` and `CERTS_SECRET_NAME`), append a default filter chain to the TLS listener (applied if none of the SNI matches).

/kind bug

Fixes #740
Initial issue (with domain mapping) here: https://github.com/knative/serving/issues/12377#issuecomment-1005019057

**Release Note**

Not sure about this, but we can go for this (open to suggestions):

```release-note
Add a filter chain to the HTTPS listener with SNI, applied if no SNI matches
```

### DIFF
--- a/pkg/generator/caches.go
+++ b/pkg/generator/caches.go
@@ -245,15 +245,15 @@ func generateListenersAndRouteConfigs(
 
 		// if a certificate is configured, add a new filter chain to TLS listener
 		if useHTTPSListenerWithOneCert() {
-			externalHTTPSEnvoyListenerWithOneCert, err := newExternalEnvoyListenerWithOneCert(
+			externalHTTPSEnvoyListenerWithOneCertFilterChain, err := newExternalEnvoyListenerWithOneCertFilterChain(
 				ctx, externalTLSManager, kubeclient,
 			)
 			if err != nil {
 				return nil, nil, err
 			}
 
-			externalHTTPSEnvoyListener.FilterChains = append(externalHTTPSEnvoyListenerWithOneCert.FilterChains,
-				externalHTTPSEnvoyListener.FilterChains...)
+			externalHTTPSEnvoyListener.FilterChains = append(externalHTTPSEnvoyListener.FilterChains,
+				externalHTTPSEnvoyListenerWithOneCertFilterChain)
 		}
 
 		listeners = append(listeners, externalHTTPSEnvoyListener)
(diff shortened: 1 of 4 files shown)


norbjd:
@googlebot I signed it!

norbjd:
I've signed the CLA with my GitHub username so the check should work now :stuck_out_tongue:

pkg/generator/caches.go
```diff
@@ -242,6 +242,20 @@ func generateListenersAndRouteConfigs(
 		if err != nil {
 			return nil, nil, err
 		}
+
+		// if a certificate is configured, add a new filter chain to TLS listener
+		if useHTTPSListenerWithOneCert() {
+			externalHTTPSEnvoyListenerWithOneCert, err := newExternalEnvoyListenerWithOneCert(
```
nak3, line 248: Although `newExternalEnvoyListenerWithOneCert()` creates a listener, we don't use the Listener but just want to use the `FilterChain`. Should we introduce a new function (or split `NewHTTPSListener`) to create the Chain?

nak3 (commented):
Thank you! LGTM overall.
Just left one comment.

nak3:
Note to self: 
Envoy docs does not explicitly mention about empty `server_names` - https://www.envoyproxy.io/docs/envoy/latest/api-v3/config/listener/v3/listener_components.proto#config-listener-v3-filterchainmatch
But it seems we can use empty `server_names` for non-SNI or unknown SNI as sper https://github.com/envoyproxy/envoy/issues/15869#issuecomment-816205965.

pkg/generator/caches.go
norbjd: Thanks for your suggestion :raised_hands:
You're right, I changed that in my last commit : https://github.com/knative-sandbox/net-kourier/pull/743/commits/81aadf13824a72321b9492e34a639d0ddfb7435b. It required some refacto, but nothing too difficult.

norbjd:
@nak3 Thanks for your kind review. I made some modifications according to it :+1: 

[Envoy docs](https://www.envoyproxy.io/docs/envoy/latest/api-v3/config/listener/v3/listener_components.proto#config-listener-v3-filterchainmatch) are actually mentioning that `server_names` can be empty (see description of `server_names` field):

> (repeated string) **If non-empty**, a list of server names (e.g. SNI for TLS protocol) to consider when determining a filter chain match.

From what I understand, "*if non-empty*" implies that *empty* is a possible value.

nak3:
/retest
/lgtm
/approve

Thank you so much @norbjd !

PR #744, opened 2022-01-10
enable proxy protocol

RamyChaabane:
# Changes

:gift: enable proxy protocol 

/kind enhancement

This patch will help users to retreive connection information from the source requesting the connection, especially the source IP address
**Release Note**

```release-note
net-kourier supports proxy protocol
action required
```

**Docs**

```docs
[README.md]: <https://github.com/RamyChaabane/net-kourier/tree/SRVLESS-7277-proxy-protocol#proxy-protocol-configuration>
```


RamyChaabane:
/assign @evankanderson

I will add some tests if this is approved, thx

evankanderson:
/assign @carlisia 

For a second set of eyes / experience

mcastilho:
Isn't this proxy support feature a must have if you are using AWS NLB ? I have everything working with Knative + Courier expect that all the request IPs are the same (the load balancer). 

Am I missing something? Has any of you able to get original IP address without this feature using a Layer 4 LB (AWS NLB in this case)

I can't wait for this to be merged in.

carlisia:
Hey @mcastilho, unless someone gets to this first, I'll review the PR tomorrow.

I haven't used net-kourier, but we do have conformance tests that run frequently for all components, although they might only be hitting GCP now that I think about it. I'll check on that.

nak3:
@RamyChaabane Thank you for your PRs. Do you have any plan to add an e2e test for this feature? If I understand correctly, we need to setup a L4 proxy (which supports proxy protocol) in front. How will you tackle it?

Also, if we support further Envoy features in Kourier (like this PR and [pull/759](https://github.com/knative-sandbox/net-kourier/pull/759)), Kourier will become Istio Gateway or Contour eventually... Frankly speaking, why you want to add the features into Kourier rather than using Istio (or Contour)? Please don't get me wrong. I am happy to improve Kourier and glad to see your PRs but I am curious about it.

RamyChaabane:
@nak3 Thank you for your comment. We've tested Istio Gateway and Contour and we've noticed issues when it comes to deployment time and even scaling, especially for Contour: it does not scale pretty well when we have reached arround 500 KSVC and the deployment was taking too much time. Adding those features does not change much to Kourier.

Regarding tests, I'm planning to add some units tests if this feature will be accepted. For e2e tests, I can do that, but we will need, as you said, to setup a L4 proxy and rerun eventually all the existing tests, so we migh need to setup another cluster(s) for that

pkg/config/config.go
```diff
@@ -38,6 +38,10 @@ const (
 	HTTPPortInternal = uint32(8081)
 	// HTTPSPortExternal is the port for external HTTPS availability.
 	HTTPSPortExternal = uint32(8443)
+	// HTTPPortProbProxyProtocol is the port for prob when proxy-protocol is enabled
```
nak3, line 41: If we use `8090` and `9443` for probe by default (i.e. net-kourier always use `8090` and `9443` ports to probe regardless proxy-protocol enabled or disabled), the code especially `lister.go`  becomes much simpler?
Do you think it is a bad idea?

config/200-config.yaml
```diff
@@ -44,3 +44,8 @@ data:
     # in the context of services should be logged. Readiness
     # probes etc. must be configured via the bootstrap config.
     enable-service-access-logging: "true"
+
+    # Specifies whether to use proxy-protocol in ordrer to to safely
+    # transport connection information such as a client's address
+    # across multiple layers of TCP proxies.
+    proxy-protocol: "false"
```
nak3, line 51: Do we want to use this one global configuration for the proxy-protcol?
Is it not better to introduce an annotation like `networking.knative.dev/proxy-protocol: true` to Ingress object? Then, I wonder we can add a few tests to verify the proxy-protocol instead of setup another cluster :thinking:

pkg/config/config.go
RamyChaabane: I think it's a good idea, I'll do that

config/200-config.yaml
RamyChaabane: If I understood correctly, we wanna have an L4 proxy that suppports proxy protocol, but headers will be forwarded only if the ksvc/ingress is asking for them ...

Unless I'm missing something, to enable proxy protocol with envoy, we should configure as a listener filter (ListenerFilter at the listener level) section and it cannot be configured by ksvc/ingress

RamyChaabane:
@nak3, I've updated the code and now ports 8090 and 9443 are the prob ports by default, can't wait for the feedback

nak3:
Thank you! And sorry I will be day off until next week but will take a look at it.

config/200-config.yaml
```diff
@@ -44,3 +44,8 @@ data:
     # in the context of services should be logged. Readiness
     # probes etc. must be configured via the bootstrap config.
     enable-service-access-logging: "true"
+
+    # Specifies whether to use proxy-protocol in ordrer to to safely
+    # transport connection information such as a client's address
+    # across multiple layers of TCP proxies.
```
nak3, line 50: Could you mention that this is an experimental/alpha feature just in case?

https://github.com/knative/serving/blob/77a0443c52fba683abd88a76e645244cb38e30ac/config/core/configmaps/deployment.yaml#L101

pkg/envoy/api/route.go
```diff
@@ -52,7 +52,7 @@ func NewRoute(name string,
 		}
 	}
 
-	return &route.Route{
+	newRoute := &route.Route{
```
nak3, line 55: Does this change something? I think we don't need to change it.

pkg/generator/caches.go
```diff
@@ -254,18 +274,30 @@ func generateListenersAndRouteConfigs(
 
 			externalHTTPSEnvoyListener.FilterChains = append(externalHTTPSEnvoyListener.FilterChains,
 				externalHTTPSEnvoyListenerWithOneCertFilterChain)
+
```
nak3, line 277: Could you drop unnecessary blank line? :sweat_smile: Not only this one, but also there are some unnecessary break lines.

nak3:
Thank you for the update.
I learned curl command has `--haproxy-protocol` option and it works against your change.
So, I guess we can add e2e test under `test/proxyproto` by using https://github.com/pires/go-proxyproto or library?

README.md
```diff
@@ -133,6 +134,71 @@ vars in the `net-kourier-controller` deployment:
 
 `*` Required
 
+## Proxy Protocol Configuration
+
```
nak3, line 138: Could you mention about the following configuration?

```shell
kubectl patch configmap/config-kourier \
  -n knative-serving \
  --type merge \
  -p '{"data":{"enable-proxy-protocol":"true"}}'
```

RamyChaabane:
@nak3 Thank you for the feedback
I updated the code according to your comments, if you feel it's ready to go, I can start add some tests

pkg/envoy/api/http_connection_manager.go
```diff
@@ -63,6 +63,8 @@ func NewHTTPConnectionManager(routeConfigName string, enableAccessLog bool) *hcm
 				RouteConfigName: routeConfigName,
 			},
 		},
+		//Force the connection manager to use the real remote address of the client connection.
+		UseRemoteAddress: &wrapperspb.BoolValue{Value: true},
```
nak3, line 67: Could you set this `use_remote_address` only when proxy-protocol is enabled? It may be a breaking change for the existing users.

nak3:
Thank you @RamyChaabane I left one more comment but I think it is ready to go. Let's add some tests.

RamyChaabane:
@nak3 I've added some units tests and now we have a good coverage, I'll start working on e2e tests once you're OK with these changes

nak3:
Thank you! Sure, it should be alright to start adding the e2e tests.

RamyChaabane:
@nak3 e2e tests has been added and passed successfully

nak3:
Thank you so much! I will look at it early next week.

nak3 (commented):
Looks good to me overall.
Leaving some nits comments.

pkg/generator/caches.go
```diff
@@ -219,26 +219,45 @@ func generateListenersAndRouteConfigs(
 	internalRouteConfig := envoy.NewRouteConfig(internalRouteConfigName, clusterLocalVirtualHosts)
 
 	// Now we setup connection managers, that reference the routeconfigs via RDS.
-	externalManager := envoy.NewHTTPConnectionManager(externalRouteConfig.Name, cfg.Kourier.EnableServiceAccessLogging)
-	externalTLSManager := envoy.NewHTTPConnectionManager(externalTLSRouteConfig.Name, cfg.Kourier.EnableServiceAccessLogging)
-	internalManager := envoy.NewHTTPConnectionManager(internalRouteConfig.Name, cfg.Kourier.EnableServiceAccessLogging)
-	externalHTTPEnvoyListener, err := envoy.NewHTTPListener(externalManager, config.HTTPPortExternal)
+	externalManager := envoy.NewHTTPConnectionManager(externalRouteConfig.Name, cfg.Kourier.EnableServiceAccessLogging, cfg.Kourier.EnableProxyProtocol)
```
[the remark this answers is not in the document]
RamyChaabane: I'll do another PR for this

nak3:
/hold cancel
/lgtm
/approve

Thank you so much for your patience and this improvement @RamyChaabane 
Let's merge this.

PR #753, opened 2022-01-14
Fix leader election issues due to defaults change

nak3:
This patch adds `test/config/leader-election.yaml` which was also added by https://github.com/knative/serving/pull/12506


nak3:
/cc @rhuss @skonto

Could you please take a look? This is blocking the auto update https://github.com/knative-sandbox/net-kourier/pull/751.

dprotaso:
Interesting - `visibility/split` errored out. I see this in net-contour and wondered if it was unique to us. Is this a kourier flake as well?

/retest

PR #756, opened 2022-01-19
Bump k8s version to 1.23.0 in kind

nak3:
As k8s min version was bumped to 1.21 by https://github.com/knative/pkg/pull/2397, this patch bumps k8s version in kind.


dprotaso (approved):
/lgtm
/approve

PR #785, opened 2022-03-06
Ext auth service : remove `split_words` to honor `KOURIER_EXTAUTHZ*` env vars

norbjd:
# Changes

:bug: Ext auth service : remove `split_words` to honor `KOURIER_EXTAUTHZ*` env vars

/kind bug

With this fix, `KOURIER_EXTAUTHZ_FAILUREMODEALLOW` and `KOURIER_EXTAUTHZ_MAXREQUESTBYTES` env vars are honored as the docs states. Before that, users should set `KOURIER_EXTAUTHZ_FAILURE_MODE_ALLOW` and `KOURIER_EXTAUTHZ_MAX_REQUEST_BYTES` (notice the separating `_`) to change those parameters (see [envconfig `split_words` documentation](https://github.com/kelseyhightower/envconfig/blob/v1.4.0/README.md#struct-tag-support)).

**Release Note**

No impacts on clients if they were following the [config of ext auth in the `README.md`](https://github.com/knative-sandbox/net-kourier/blob/a95fd30/README.md#external-authorization-configuration).

**Docs**

No docs.

### DIFF
--- a/pkg/config/ext_authz.go
+++ b/pkg/config/ext_authz.go
@@ -54,10 +54,10 @@ type ExternalAuthzConfig struct {
 }
 
 type config struct {
-	Host             string `split_words:"true"`
-	FailureModeAllow bool   `split_words:"true"`
-	MaxRequestBytes  uint32 `split_words:"true" default:"8192"`
-	Timeout          int    `split_words:"true" default:"2000"`
+	Host             string
+	FailureModeAllow bool
+	MaxRequestBytes  uint32 `default:"8192"`
+	Timeout          int    `default:"2000"`
 }
 
 func init() {


nak3 (changes requested):
I think this is a docs bug and we should fix README rather than code.

> No impacts on clients if they were following the [config of ext auth in the README.md](https://github.com/knative-sandbox/net-kourier/blob/a95fd30/README.md#external-authorization-configuration).

If they were following the README.md, the configuration does not work, does it?

norbjd:
Hello @nak3 !

As of now, if users are following this part in the README.md:

> KOURIER_EXTAUTHZ_FAILUREMODEALLOW*: Allow traffic to go through if the ext auth service is down. Accepts true/false
> KOURIER_EXTAUTHZ_MAXREQUESTBYTES: Max request bytes, if not set, defaults to 8192 Bytes. More info [Envoy Docs](https://www.envoyproxy.io/docs/envoy/latest/api-v3/extensions/filters/http/ext_authz/v3/ext_authz.proto.html?highlight=max_request_bytes#extensions-filters-http-ext-authz-v3-buffersettings)

Then their parameters are not used in Kourier. But there is no error, so users could think that they have correctly change the parameters.

To be taken into account by Kourier, they have to set instead:

- `KOURIER_EXTAUTHZ_FAILURE_MODE_ALLOW`
- `KOURIER_EXTAUTHZ_MAX_REQUEST_BYTES`

which is not documented anywhere. They have to dig into the code to find those values.

I have changed the actual behavior instead of the docs, because I don't want this to be a breaking change for former users of `KOURIER_EXTAUTHZ_FAILUREMODEALLOW` and `KOURIER_EXTAUTHZ_MAXREQUESTBYTES`. To my opinion (but I might be wrong), people tend to follow the docs instead of diving into the source code.

Anyway, both approaches are ok, so tell me the approach you prefer! Note that if we change the docs, we have to inform users using these env vars to change them (add underscores).

nak3:
/ok-to-test

Thank you for your explanation. 
In addition to your comment, as per this commit history https://github.com/knative-sandbox/net-kourier/commit/2926c6a4461c3cf087face6c8b31c4cf3e5e1eec, `KOURIER_EXTAUTHZ_FAILUREMODEALLOW` and `KOURIER_EXTAUTHZ_MAXREQUESTBYTES` were originally expected (you might already know :sweat_smile: ), so let's use your change.

nak3:
@norbjd Just in case, is it alright to merge this after v1.3 release? Today is the [v1.3 release day](https://github.com/knative/release#schedule) (I think it will start when US folks are on-line) so I am hesitate to merge this just before the release.

norbjd:
That's ok for me @nak3 ! Let's merge this after the v1.3.0 release :wink:

Issue #750, opened 2022-01-13
All Hops Encrypted: alpha Kourier support for encrypted backends


evankanderson:
Larger description in the [Feature Track document](https://docs.google.com/document/d/1XE7UzgQlVVtAb7ULSqOyKCaIHtm8zMF35ainp1JmwyY/edit?resourcekey=0-e1MXRxQaalq-EHW46ZCCWg#heading=h.n8a530nnrb)

# Summary:

Kourier should support calling activator / backends with a known CA key and subject name (provided by the cluster administrator in `config-network` for the All Hops Encrypted alpha).

Expected `config-network` keys:

* `activator-ca` -- contains the CA public certificate used to sign the activator TLS certificate
* `activator-name` -- contains the SAN (Subject Alt Name) used to validate the activator TLS certificate

This probably involves extending the [`Cluster` configuration](https://www.envoyproxy.io/docs/envoy/latest/api-v3/config/cluster/v3/cluster.proto#config-cluster-v3-cluster) with an [`UpstreamTlsContext`](https://www.envoyproxy.io/docs/envoy/latest/api-v3/extensions/transport_sockets/tls/v3/tls.proto#extension-envoy-transport-sockets-tls) and [`CommonTlsContext`](https://www.envoyproxy.io/docs/envoy/latest/api-v3/extensions/transport_sockets/tls/v3/tls.proto#extension-envoy-transport-sockets-tls), and possibly implementing SDS for these certificates.

/kind feature-request

evankanderson:
Ref https://github.com/nak3/net-kourier/pull/11 for a prototype

PR #795, opened 2022-03-15
Watch config-network configmap

nak3:
Currently net-kourier controller does not watch `config-network` configmap as it does not use any configuration.

Now, https://github.com/knative-sandbox/net-kourier/issues/750 introduced `activator-ca` and `activator-san(name)` so controller needs to watch the configmap.

This patch adds it.

Part of: https://github.com/knative-sandbox/net-kourier/pull/761


nak3:
Closing as https://github.com/knative-sandbox/net-kourier/pull/761 was merged.

PR #802, opened 2022-03-22
Fix coverage settings

skonto:
# Changes
Fixes the coverage ci settings as discussed [here](https://github.com/knative-sandbox/net-kourier/pull/761#issuecomment-1072264476).

/hold for #761

### DIFF
--- a/.codecov.yaml
+++ b/.codecov.yaml
@@ -13,6 +13,7 @@ coverage:
       default: false
 ignore:
   - "**/zz_generated*.go" # Ignore generated files.
+  - "**/zz_generated.deepcopy.go" # Ignore generated files.
   - "**/*.pb.go" # Ignore proto-generated files.
   - "hack"
   - "pkg/client"


nak3:
Hmm...
The `.codecov.yaml` is used by github action  https://github.com/knative-sandbox/net-kourier/blob/main/.github/workflows/knative-go-test.yaml#L66-L70 but is it true that `pull-knative-sandbox-net-kourier-go-coverage`also use it? 
Please note that the `go-coverage`'s code is here https://github.com/knative/test-infra/tree/main/tools/coverage/.

Also, if this is proper fix, we need to apply all repos like https://github.com/knative/serving/blob/main/.codecov.yaml#L18 https://github.com/knative/eventing/blob/main/.codecov.yaml#L18 https://github.com/knative/pkg/blob/main/.codecov.yaml#L20.
So if you don't mind, could you bring up your fix against serving or pkg repo? I would like net-kourier repo to follow the parent repos.

nak3 (changes requested):
Would like to understand https://github.com/knative-sandbox/net-kourier/pull/802#issuecomment-1075181763

skonto:
@nak3 already merged in pkg https://github.com/knative/pkg/pull/2471. Will do for Serving too. Not sure how this file is being picked up by prow, I could dig into it.

nak3:
> @nak3 already merged in pkg https://github.com/knative/pkg/pull/2471. 

I mean why the change `- "**/zz_generated.deepcopy.go" # Ignore generated files.` needs only for this repository? You didn't change it in pkg repo.

> Not sure how this file is being picked up by prow, I could dig into it.

Yes, please. I still cannot believe that the prow uses `.codecov.yaml`.

skonto:
@nak3 I did ignore whole folders in pkg so no need for the pattern.I dont have access to see the actual reports as the ones failing in #824

nak3:
I don't think you ignore whole folders in pkg repo - for example https://github.com/knative/pkg/blob/main/apis/zz_generated.deepcopy.go.

And again, I still cannot believe that the prow uses `.codecov.yaml`...

skonto:
Ok will try find what prow does as stated before but no promise.

skonto:
Given my comment [here](https://github.com/knative-sandbox/net-kourier/pull/824#issuecomment-1100025031) we can close this PR. The file is indeed used but previous failures were not due to the file pattern.

nak3:
Well... You tried to fix this prow issue https://github.com/knative-sandbox/net-kourier/pull/761#issuecomment-1075810320, right?
You started seeing different coverage.

nak3:
> but it seems the coverage tooling has changed.

No.... The prow produces the coverage for the file which was only changed. So, the issue does not happen as long as `zz_generated.deepcopy.go` was not changed.

Anyway, this PR is not correct and the fix should be sent against test-infra repo. Closing.

skonto:
> Please note that the go-coverage's code is here https://github.com/knative/test-infra/tree/main/tools/coverage/.
This has moved.

PR #813, opened 2022-03-30
Use HTTP2 when service has HTTP2 port in the ports

nak3:
https://github.com/knative/serving/pull/12770 will change the service to have multiple ports like this:

```
$ kubectl  get svc 
NAME                          TYPE           CLUSTER-IP     EXTERNAL-IP                                         PORT(S)                                              AGE
hello-example                 ExternalName   <none>         kourier-internal.kourier-system.svc.cluster.local   80/TCP                                               5h39m
hello-example-00001           ClusterIP      10.96.16.165   <none>                                              80/TCP,443/TCP                                       5h40m
hello-example-00001-private   ClusterIP      10.96.11.168   <none>                                              80/TCP,443/TCP,9090/TCP,9091/TCP,8022/TCP,8012/TCP   5h40m
```

and KIngress:

```
$ kubectl  get king  -o yaml
  ..
  spec:
    - hosts:
      - hello-example.default.example.com
      http:
        paths:
        - splits:
            serviceName: hello-example-00001
            serviceNamespace: default
            servicePort: 443
      visibility: ExternalIP
```

So, Kourier misses to use HTTP2 when the servicePort `443`. This patch fixes it.

### DIFF
--- a/pkg/generator/ingress_translator_test.go
+++ b/pkg/generator/ingress_translator_test.go
@@ -891,6 +891,73 @@ func TestIngressTranslatorWithUpstreamTLS(t *testing.T) {
 				internalVirtualHosts:    vHosts,
 			}
 		}(),
+	}, {
+		name: "http2 and https",
+		in: ing("simplens", "simplename", func(ing *v1alpha1.Ingress) {
+			ing.Spec.Rules[0].HTTP.Paths[0].RewriteHost = ""
+			ing.Spec.Rules[0].HTTP.Paths[0].Splits[0].ServicePort = intstr.FromString("https")
+		}),
+		state: []runtime.Object{
+			svc("servicens", "servicename", func(service *corev1.Service) {
+				service.Spec.Ports = []corev1.ServicePort{{
+					Name:       "http2",
+					TargetPort: intstr.FromInt(8080),
+				}, {
+					Name:       "https",
+					TargetPort: intstr.FromInt(443),
+				}}
+			}),
+			eps("servicens", "servicename"),
+			caSecret,
+		},
+		want: func() *translatedIngress {
+			vHosts := []*route.VirtualHost{
+				envoy.NewVirtualHost(
+					"(simplens/simplename).Rules[0]",
+					[]string{"foo.example.com", "foo.example.com:*"},
+					[]*route.Route{envoy.NewRoute(
+						"(simplens/simplename).Rules[0].Paths[/test]",
+						[]*route.HeaderMatcher{{
+							Name: "testheader",
+							HeaderMatchSpecifier: &route.HeaderMatcher_ExactMatch{
+								ExactMatch: "foo",
+							},
+						}},
+						"/test",
+						[]*route.WeightedCluster_ClusterWeight{
+							envoy.NewWeightedCluster("servicens/servicename", 100, map[string]string{"baz": "gna"}),
+						},
+						0,
+						map[string]string{"foo": "bar"},
+						""),
+					},
+				),
+			}
+
+			return &translatedIngress{
+				name: types.NamespacedName{
+					Namespace: "simplens",
+					Name:      "simplename",
+				},
+				sniMatches: []*envoy.SNIMatch{},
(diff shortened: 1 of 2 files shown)


nak3:
@skonto  @rhuss  bump

pkg/generator/ingress_translator_test.go
```diff
@@ -891,6 +891,73 @@ func TestIngressTranslatorWithUpstreamTLS(t *testing.T) {
 				internalVirtualHosts:    vHosts,
 			}
 		}(),
+	}, {
+		name: "http2 and https",
```
skonto, line 895: The test passes even if we do:

```
for _, port := range service.Spec.Ports {
  if port.Port == split.ServicePort.IntVal || port.Name == split.ServicePort.StrVal {
    externalPort = port.Port
    targetPort = port.TargetPort.IntVal
  }
  if port.Name == "http2" || port.Name == "h2c" {
    http2 = true
  }
}
```
Assignment `http2 = port.Name == "http2" || port.Name == "h2c"` was problematic because it would override the previous value.
Is there a scenario where:
`if port.Port != split.ServicePort.IntVal && port.Name != split.ServicePort.StrVal`
and `if port.Name == "http2" || port.Name == "h2c"`?

nak3: The logic is one of service port has `http2` port, we use `http2`. So it is not problematic.

skonto: > The logic is that one of service port has http2 port, then we use http2. So it is not problematic.
Yes I understand that, but since the test passes we should keep the logic stricter and not move that outside the loop.
My question still holds what is the gain of moving that part outside the first condition?

nak3: Okay, added one more test case for "http and https".

nak3: @skonto If you have some suggestion, could you please provide the code snippet?

skonto: > When the service has two ports (e.g. http2 and https) but Ingress has one port 443, the scenario you mentioned happens.

Do you have a test case for that? If the test case is `name: "http2 and https"`, then as I mentioned above having the second within the first still makes the test pass:
```
for _, port := range service.Spec.Ports {
  if port.Port == split.ServicePort.IntVal || port.Name == split.ServicePort.StrVal {
    externalPort = port.Port
    targetPort = port.TargetPort.IntVal
  }
  if port.Name == "http2" || port.Name == "h2c" {
    http2 = true
  }
}
```
So I dont think we test then the case where `port.Port != split.ServicePort.IntVal && port.Name != split.ServicePort.StrVal` is true. Reason is port.Port == split.ServicePort.IntVal is always true in this test. Both equal to zero.

skonto: A scenario where you actually test this is:
```
		state: []runtime.Object{
			svc("servicens", "servicename", func(service *corev1.Service) {
				service.Spec.Ports = []corev1.ServicePort{{
					Name:       "http2",
					TargetPort: intstr.FromInt(8080),
					**Port: 80,**
				}, {
					Name:       "https",
					TargetPort: intstr.FromInt(443),
				}}
			}),
			eps("servicens", "servicename"),
			caSecret,
		},
```

nak3: Okay, so you wanted to suggest the missing port in the test case. Thank you. Updated.
And the snippet in the initial comment https://github.com/knative-sandbox/net-kourier/pull/813#discussion_r840440109, I guess it is a mistake and you probably wanted to wirte the snippet in https://github.com/knative-sandbox/net-kourier/pull/813#discussion_r840470277 :thinking:

skonto: Yes :) Sorry it was a bit hard to explain this. :+1:

nak3: Nah, it's my bad. Thank you for the review :bow:

Issue #805, opened 2022-03-23
All Hops Encrypted: alpha Kourier support for encrypted for internal traffic


nak3:
# Summary:

https://github.com/knative-sandbox/net-kourier/issues/750 will fix a traffic (`kourier` service) but the internal traffic against `kourier-internal` is still not supported.

The traffic against `hello.example.svc.cluster.local:443` from another pod should be encrypted.

/kind feature-request

PR #824, opened 2022-04-13
All Hops Encrypted: alpha Kourier support for encrypted for internal traffic

nak3:
Fix https://github.com/knative-sandbox/net-kourier/issues/805

As described in https://github.com/knative-sandbox/net-kourier/issues/805, service to service traffic does not
support TLS encryption. `kourier-internal` service does not even open the port.

This patch fixes it. 

/cc @skonto @rhuss

### DIFF
--- a/config/200-config.yaml
+++ b/config/200-config.yaml
@@ -51,9 +51,10 @@ data:
     enable-proxy-protocol: "false"
 
     # The server certificates to serve the internal TLS traffic for Kourier Gateway.
-    # It is specified by the secret name, which has the "tls.crt" and "tls.key" data field.
+    # It is specified by the secret name in controller namespace, which has
+    # the "tls.crt" and "tls.key" data field.
     # Use an empty value to disable the feature (default).
     #
     # NOTE: This flag is in an alpha state and is mostly here to enable internal testing
     #       for now. Use with caution.
-    kourier-internal-cert-secret: ""
+    cluster-cert-secret: ""
--- a/config/200-config.yaml
+++ b/config/200-config.yaml
@@ -49,3 +49,11 @@ data:
     # across multiple layers of TCP proxies.
     # NOTE THAT THIS IS AN EXPERIMENTAL / ALPHA FEATURE
     enable-proxy-protocol: "false"
+
+    # The server certificates to serve the internal TLS traffic for Kourier Gateway.
+    # It is specified by the secret name, which has the "tls.crt" and "tls.key" data field.
+    # Use an empty value to disable the feature (default).
+    #
+    # NOTE: This flag is in an alpha state and is mostly here to enable internal testing
+    #       for now. Use with caution.
+    kourier-internal-cert-secret: ""
--- a/pkg/config/configmap.go
+++ b/pkg/config/configmap.go
@@ -33,15 +33,15 @@ const (
 	// enableProxyProtocol is the config map key for enabling proxy protocol
 	enableProxyProtocol = "enable-proxy-protocol"
 
-	// kourierInternalCertSecret is the config map key for kourier internal certificates
-	kourierInternalCertSecret = "kourier-internal-cert-secret"
+	// clusterCertSecret is the config map key for kourier internal certificates
+	clusterCertSecret = "cluster-cert-secret"
 )
 
 func DefaultConfig() *Kourier {
 	return &Kourier{
 		EnableServiceAccessLogging: true, // true is the default for backwards-compat
 		EnableProxyProtocol:        false,
-		KourierInternalCertSecret:  "",
+		ClusterCertSecret:          "",
 	}
 }
 
@@ -52,7 +52,7 @@ func NewConfigFromMap(configMap map[string]string) (*Kourier, error) {
 	if err := cm.Parse(configMap,
 		cm.AsBool(enableServiceAccessLoggingKey, &nc.EnableServiceAccessLogging),
 		cm.AsBool(enableProxyProtocol, &nc.EnableProxyProtocol),
-		cm.AsString(kourierInternalCertSecret, &nc.KourierInternalCertSecret),
+		cm.AsString(clusterCertSecret, &nc.ClusterCertSecret),
 	); err != nil {
 		return nil, err
 	}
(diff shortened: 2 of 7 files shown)


nak3:
@evankanderson https://github.com/knative/serving/issues/11906 did not mention about the internal traffic (`*.svc.cluster.local`) but our customer requested that "all" traffic so we need to add this. Please let us know if you think we should support it in `config-network` rather than Kourier specific.

pkg/generator/caches.go
```diff
@@ -336,3 +367,12 @@ func newExternalEnvoyListenerWithOneCert(ctx context.Context, manager *httpconnm
 
 	return envoy.NewHTTPSListener(config.HTTPSPortExternal, []*v3.FilterChain{filterChain}, enableProxyProtocol)
 }
+
+func newInternalEnvoyListenerWithOneCert(ctx context.Context, manager *httpconnmanagerv3.HttpConnectionManager, kubeClient kubeclient.Interface, enableProxyProtocol bool, certSecretName string) (*v3.Listener, error) {
```
skonto, line 371: The only diff between internal and external is the namespace  and secretaName pair? 
```
os.Getenv(envCertsSecretNamespace) , os.Getenv(envCertsSecretName)
vs
ystem.Namespace(), certSecretName,
``` 
Should we simplify this?

skonto:
@nak3 for the coverage reports I dont have access to evaluate what is wrong, obviously some files liek config.go could have more coverage. I left a comment in #802

evankanderson:
How does this work with Addressable resources? Will they refuse to connect without trusting the (self-signed?) CA for the `*.svc.cluster.local` address?

I'm concerned about unintended side effects of skipping this into the 1.4 build with only 3-4 business days of testing before the release. If we are going to do it, it should definitely be confined to `net-kourier` so that side effects can be mitigated by switching specific services to other ingress implementations.

config/200-config.yaml
```diff
@@ -49,3 +49,11 @@ data:
     # across multiple layers of TCP proxies.
     # NOTE THAT THIS IS AN EXPERIMENTAL / ALPHA FEATURE
     enable-proxy-protocol: "false"
+
+    # The server certificates to serve the internal TLS traffic for Kourier Gateway.
+    # It is specified by the secret name, which has the "tls.crt" and "tls.key" data field.
+    # Use an empty value to disable the feature (default).
+    #
```
evankanderson, line 56: This is resolved in the kourier namespace, correct?

pkg/config/configmap.go
```diff
@@ -32,12 +32,16 @@ const (
 
 	// enableProxyProtocol is the config map key for enabling proxy protocol
 	enableProxyProtocol = "enable-proxy-protocol"
+
+	// kourierInternalCertSecret is the config map key for kourier internal certificates
+	kourierInternalCertSecret = "kourier-internal-cert-secret"
```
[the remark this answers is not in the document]
evankanderson: This is being used for (client on the cluster) -> `kourier/envoy` IIUC.

config/200-config.yaml
```diff
@@ -49,3 +49,11 @@ data:
...
+    # NOTE: This flag is in an alpha state and is mostly here to enable internal testing
+    #       for now. Use with caution.
+    kourier-internal-cert-secret: ""
```
evankanderson, line 59: Suggest calling this `cluster` rather than `internal` -- it's a bit clearer about what it's for.

evankanderson: Also, I don't think you need the `kourier-` prefix, since this is in `config-kourier` already.

pkg/config/configmap.go
```diff
@@ -68,4 +73,7 @@ type Kourier struct {
 	EnableServiceAccessLogging bool
 	// EnableProxyProtocol specifies whether proxy protocol feature is enabled
 	EnableProxyProtocol bool
+	// KourierInternalCertSecret specifies the secret name for the server certificates of
+	// Kourier Internal.
+	KourierInternalCertSecret string
```
evankanderson, line 78: Again, I'd call this `ClusterCertSecret`, so as not to repeat ("stutter") `Kourier`, and to be more clear that this is a cert that clients in the cluster will see.

pkg/generator/caches.go
evankanderson: Alternatively, rather than making the cert name configurable, you could hard code it, and use the presence / absence of the cert to activate the encryption on the internal interface.

In my experience, this can work fairly well if the component is installed in its own namespace and isn't expected to share. In that case, you could default the secret name to e.g. `kourier-internal-domain-cert` or the like, and then document that if that cert is created, Kourier will use it.

evankanderson (commented):
Reading through this, it looks like it doesn't disable HTTP serving, nor does it change the published URL of the KIngress. If that's true (I didn't try running it, and I'm only seeing the changed files, not the final outcome), then I'm less worried about the short-term impact on Addressable.

pkg/generator/caches.go
nak3: Hard cord name is fine, but I assumed that the future plan as:

step-1. Create the secret(=certs) automatically by cert-manager or the library in `knative.dev/pkg`.
step-2. Knative always use TLS traffic with the auto generated certs.

At step-2, `cluster-cert-secrete` in the ConfigMap starts having the default value like:

```
kind: ConfigMap
  name: config-kourier
    cluster-cert-secret: "kourier-internal-domain-cert"
```

And if users want to use their own certs rather than auto-gen certs, they can set the different value in `cluster-cert-secret`.

skonto:
@nak3 the coverage report is [here](https://codecov.io/gh/knative-sandbox/net-kourier/compare/483cd470dc95d03d63c9c853877d8df104d946e6...4a5366996df81b2524357458d1bb92bc14272dbe#).
The files marked with red are the offending ones, need better coverage.

nak3:
The coverage https://codecov.io/gh/knative-sandbox/net-kourier/compare/483cd470dc95d03d63c9c853877d8df104d946e6...4a5366996df81b2524357458d1bb92bc14272dbe# compared with non-related old commits. Just rebased and passed codeconv.

evankanderson (commented):
/lgtm
/hold cancel

PR #825, opened 2022-04-14
Bump k8s version in kind

nak3:
As per title, this patch bumps k8s version.

Fix https://github.com/knative-sandbox/net-kourier/pull/818


psschwei (commented):
/lgtm

PR #852, opened 2022-06-01
Isolate private ksvc using different envoy listener ports

lionelvillard:
Add support for multiple private ingresses backed by multiple envoy listeners (each with a unique port). 

This is an experimental Kourier-only feature guarded by a flag in `config-kourier`:

```
# Control the desired level of incoming traffic isolation.
#
# When set to the empty value (default), all incoming traffic flows through
# a shared ingress and listeners.
#
# When set to "port", incoming traffic is isolated by using different
# listener ports.
#
# NOTE: This flag is in an alpha state.
traffic-isolation: ""
```

The selection of which envoy listener to use is done at the namespace-level. The traffic of all internal services within the same namespace is redirected to the same envoy listener. Here are the annotation for mapping a namespace to a particular envoy listener:

```go

// ListenerPortAnnotationKey is the annotation key for assigning the ingress to a particular
// envoy listener port. Only applicable to internal services.
ListenerPortAnnotationKey = "kourier.knative.dev/listener-port"
```

The allocation/assignment of a namespace to an envoy listener is an orthogonal concern and not address in this PR. This can be done by manually editing namespace annotations or automatically by a Tenant controller. 

Each envoy listener is exposed using a k8s service (e.g `kourier-isolation-9000`). This PR does not automatically create/delete these services since this concern is also related to tenant management. 

Additional context:
- [Feature Track](https://docs.google.com/document/d/11hViPJlmI4ftcZVkT8x6Au4cCxRncsEq7M6UCd_UKJ8/edit)
- [PoC](https://github.com/lionelvillard/mknative/tree/main/) showing how Network Policies with this PR can be used to isolate tenants.

# Changes

- Add a new configuration parameter for enabling port-level internal services isolation 
- Generate an envoy configuration containing multiple internal services
-

/kind enhancement

Fixes #

**Release Note**

```release-note
Add a new configuration parameter for enabling port-level internal services isolation. See documentation for more details.
```

**Docs**

```docs

```

Remaining tasks are tracked in [this](https://github.com/knative-sandbox/net-kourier/issues/876) issue.

/cc @evankanderson @nak3 @rhuss


skonto:
@lionelvillard hi, could you add a link to the ft doc in the description?

lionelvillard:
> @lionelvillard hi, could you add a link to the ft doc in the description?

done.

skonto:
@lionelvillard tests seem passing will take a look at the PR and review it. /cc @nak3 pls review too.

pkg/generator/ingress_translator.go
```diff
@@ -259,12 +265,37 @@ func (translator *IngressTranslator) translateIngress(ctx context.Context, ingre
 			}
 		}
 	}
+	listener := ""
+	port := uint32(0)
+
+	if config.FromContextOrDefaults(ctx).Kourier.TrafficIsolation == pkgconfig.IsolationIngressPort {
+		logger.Infof("Getting namespace %v\n", ingress.Namespace)
```
evankanderson, line 272: Do we still need this `logger`?

pkg/reconciler/ingress/controller.go
```diff
@@ -182,6 +183,9 @@ func NewController(ctx context.Context, cmw configmap.Watcher) *controller.Impl
 		func(ns, name string) (*corev1.Service, error) {
 			return serviceInformer.Lister().Services(ns).Get(name)
 		},
+		func(name string) (*corev1.Namespace, error) {
+			return kubernetesClient.CoreV1().Namespaces().Get(context.TODO(), name, metav1.GetOptions{})
```
evankanderson, line 187: Use the `ctx` from `NewController`, rather than `TODO`?

pkg/reconciler/ingress/ingress.go
```diff
@@ -77,6 +81,18 @@ func (r *Reconciler) ReconcileKind(ctx context.Context, ing *v1alpha1.Ingress) r
 		}
 		if ready {
 			external, internal := config.ServiceHostnames()
+
+			ns, err := r.kubernetesClient.CoreV1().Namespaces().Get(ctx, ing.Namespace, metav1.GetOptions{})
```
evankanderson, line 85: Is this going to go through the informer cache this way? I'd expect to see a `NamespacesLister` injected here.

pkg/reconciler/ingress/lister.go
```diff
@@ -82,9 +86,21 @@ func (l *gatewayPodTargetLister) getIngressUrls(ing *v1alpha1.Ingress, gatewayIp
 				target.URLs = domainsToURL(domains, scheme)
 			}
 		} else {
+			ns, err := l.kubeclient.CoreV1().Namespaces().Get(context.TODO(), ing.Namespace, metav1.GetOptions{})
```
evankanderson, line 89: Again, concerned this might be doing a GET to the apiserver for each probe, rather than caching the namespace annotations. If it's not, that's great. If it is, it seems like we should pass a NamespacesLister like the EndpointsLister.

evankanderson (commented):
A few small concerns about informer caches, but this generally looks good to me.

pkg/generator/caches.go
```diff
@@ -136,8 +141,20 @@ func (caches *Caches) ToEnvoySnapshot(ctx context.Context) (cache.Snapshot, erro
 	externalTLSVHosts := make([]*route.VirtualHost, 0, len(caches.translatedIngresses))
 	snis := sniMatches{}
 
+	localVHostsPerListener := make(map[string]portVHost)
+
 	for _, translatedIngress := range caches.translatedIngresses {
-		localVHosts = append(localVHosts, translatedIngress.internalVirtualHosts...)
+		if translatedIngress.listener != "" {
```
skonto, line 147: I suspect we dont support going back and forth from `port` to `none` wrt isolation.
I am thinking of a scenario where existing ingress have this part set but the feature is not on.

pkg/reconciler/ingress/controller.go
lionelvillard: maybe, though it's not the right context. Hypothetically the reconciler can create a cancelable context, call cancel which won't stop the inflight Get request. 

That said I changed the code to propagate the right context!

Thanks!

pkg/reconciler/ingress/ingress.go
lionelvillard: IIRC this is a direct call. I originally wrote the code to use a `NamespacesLister/Informer` in order to watch for namespace changes but that's not needed. The advantage of a direct call is that there is no need for a cache (saving a tiny bit of memory), as the expense of additional http requests (but no watch).

I'm happy to revert to back to using `NamespacesLister` if you feel it's a better compromise.

pkg/reconciler/ingress/lister.go
lionelvillard: Agree make sense in this context. Updating...

pkg/reconciler/ingress/ingress.go
lionelvillard: see my comment below. I will use an informer.

skonto:
@lionelvillard any plan to document policy setup. For example there is an image with all the policies but it would be nice to elaborate on it. Example what is the minimum setup per scenario.  Running the POC locally I see for example in order to make ping-pong work in the same tenant and ns and to be able to access it externally to the cluster I need to set:
```
$ oc get networkpolicy -n tenant1-ns1
NAME                               POD-SELECTOR   AGE
allow-all-from-to-tenant-ns        <none>         55s
allow-all-in-ns                    <none>         3m3s
allow-serving-to-tenant            <none>         16s
allow-tenant-ingress-from-tenant   <none>         9m13s
```
```
$ oc logs ping-00001-deployment-6574dcc4c8-k45vp  -n tenant1-ns1 -c user-container | tail
2022/06/24 17:25:51 ping succeeded
2022/06/24 17:25:53 ping succeeded
2022/06/24 17:25:55 ping succeeded
2022/06/24 17:25:57 ping succeeded
2022/06/24 17:25:59 ping succeeded
2022/06/24 17:26:01 ping succeeded
2022/06/24 17:26:03 ping succeeded
2022/06/24 17:26:05 ping succeeded
2022/06/24 17:26:07 ping succeeded
2022/06/24 17:26:09 ping succeeded
$ oc logs ping-00001-deployment-6574dcc4c8-k45vp  -n tenant1-ns1 -c user-container | tai^C
[stavros@easy mknative]$ oc logs pong-00001-deployment-79d87c75b-7n6t6    -n tenant1-ns1 -c user-container | tail
2022/06/24 17:25:59 pong
2022/06/24 17:26:01 pong
2022/06/24 17:26:03 pong
2022/06/24 17:26:05 pong
2022/06/24 17:26:07 pong
2022/06/24 17:26:09 pong
2022/06/24 17:26:11 pong
2022/06/24 17:26:13 pong
2022/06/24 17:26:15 pong
2022/06/24 17:26:17 pong
$ curl -v  http://ping.tenant1-ns1.127.0.0.1.sslip.io
*   Trying 127.0.0.1:80...
* Connected to ping.tenant1-ns1.127.0.0.1.sslip.io (127.0.0.1) port 80 (#0)
> GET / HTTP/1.1
> Host: ping.tenant1-ns1.127.0.0.1.sslip.io
> User-Agent: curl/7.71.1
> Accept: */*
> 
* Mark bundle as not supporting multiuse
< HTTP/1.1 200 OK
< content-length: 0
< date: Fri, 24 Jun 2022 17:26:23 GMT
< x-envoy-upstream-service-time: 3
< server: envoy
< 
* Connection #0 to host ping.tenant1-ns1.127.0.0.1.sslip.io left intact
```

lionelvillard:
@skonto 

> any plan to document policy setup sooner?

yes I can work on that. To make sure we are on the same page, you are referring to adding a new page in knative.dev describing a minimum network policy setup for multi-tenant environment?

> For example there is an image with all the policies but it would be nice to elaborate on it

Sorry I'm not parsing this. What image?

evankanderson:
> @skonto 
> 
> > For example there is an image with all the policies but it would be nice to elaborate on it
> 
> Sorry I'm not parsing this. What image?

Picture, not OCI artifact.

skonto:
@evankanderson :P

pkg/generator/ingress_translator.go
lionelvillard: done.

pkg/reconciler/ingress/controller.go
lionelvillard: I changed to use a informer, no NA anymore.

lionelvillard:
@skonto @nak3 @evankanderson I improved a bit the [PoC](https://github.com/lionelvillard/mknative/tree/main/) readme. It's now quite clear (I hope) what cluster admins need to do create/delete a tenant. 

Remaining tasks are tracked by these issues:
- Documentation: https://github.com/knative/docs/issues/5079 
- e2e tests: https://github.com/knative-sandbox/net-kourier/issues/870
- Internal encryption: https://github.com/knative-sandbox/net-kourier/issues/871


Are you ok with merging this PR before 1.6 is cut next week?

nak3:
I have two comments about test and annotations.

#### 1. testing

I see that there is a plan to add e2e test https://github.com/knative-sandbox/net-kourier/issues/870 but we should try to run serving's e2e to make sure if this does not break anything like this https://github.com/knative/serving/pull/13058, should't we?

#### 2. annotations in the user namesspace

I think that we should change the following annotations to a __label__ like `adimin.knative.dev/tenant-name` to stop exposing the internal information in the user's namespacee.

> `kourier.knative.dev/listener`: the envoy listener suffix to use
> `kourier.knative.dev/listener-port`: the envoy listener port

There are some reasons for that such as:
- It increases risks to atack/invade Kourier Gateway by cheating the exposed annotation.
- For troubleshooting/admin friendly, we should use `label` or at least one annotation.
- The general label instead the Kourier specific annotation allows other ingress to use the same label even if they apply the different implementation.

I guess it is possible by introducing a ConfigMap or new CRD for it.

lionelvillard:
Thanks @nak3 for your review!

> I have two comments about test and annotations.
> 
> #### 1. testing
> I see that there is a plan to add e2e test #870 but we should try to run serving's e2e to make sure if this does not break anything like this [knative/serving#13058](https://github.com/knative/serving/pull/13058), should't we?
> 

I'm on it!

> #### 2. annotations in the user namesspace
> I think that we should change the following annotations to a **label** like `adimin.knative.dev/tenant-name` to stop exposing the internal information in the user's namespacee.

> 
> > `kourier.knative.dev/listener`: the envoy listener suffix to use
> > `kourier.knative.dev/listener-port`: the envoy listener port
> 
> There are some reasons for that such as:
> 
> * It makes some risks to attack/invade Kourier Gateway by cheating the exposed annotation.

How using a label different from using an annotation? Is it possible to restrict access to labels and not annotations? 

> * For troubleshooting/admin friendly, we should use `label` or at least one annotation.
> * The general label instead of the Kourier specific annotation allows other ingress to use the same label even if they apply the different implementation.

My only concern is that there is no concept of tenant in Knative (AFAIK).

> 
> I guess it is possible by introducing a ConfigMap or new CRD for it.

nak3:
> How using a label different from using an annotation? Is it possible to restrict access to labels and not annotations?

Changing annotation to label is trivial thing. I wanted to suggest we should stop exposing the internal information by changing `adimin.knative.dev/tenant-name` or some general name.

The annotation with `adimin.knative.dev/tenant-name` is also fine, but admin can list the namespace with `kubectl  get ns  -l adimin.knative.dev/tenant-name=my-tenant` if we use the label instead of annotation, that's the only thing for annotation vs label.

lionelvillard:
@nak3 What about `kourier.knative.dev/listener-port`. Should we keep it the same or rename to `admin.knative.dev/tenant-port`?

lionelvillard:
Knative serving test PR: https://github.com/knative/serving/pull/13094

nak3:
Umm... I was thinking that we can introduce a new CRD or CM like this. Then, we can completely disallow users to change the listener name and port.

```
               +--------------------+
               | kourier controller |
               +--------------------+
          	         |
                         | read listener name & port.
 new CRD (or CM)         |
+-------------------+    |
|tenant:            | <--+
|  name: my-tenant  |
|    lister: ten-a  |
|    port: 8081     |
|    members:       |                    namespace (Kourier does not need to touch.)
|      - namespaceA |                  +------------+
|      - namespaceB |                  | namespaceA |
+-------------------+                  +------------+
```

It is not good idea?

lionelvillard:
> 



> Umm... I was thinking that we can introduce a new CRD or CM like this. Then, we can completely disallow users to change the listener name and port.

@rhuss @skonto @evankanderson WDYT?

nak3:
The most important point for me is that it is unsafe to add the `kourier.knative.dev/listener{-port}` label, which is editable, in users' namespace.
I think users can invade other tenants, other Envoy listener, etc.

(On the other hand, using a new CRD or CM is safe, one place configuration, admin friendly etc :sunglasses: .)

skonto:
I think it is safer if we rely on some crd or cm but we dont have to do it now imho. We can get this in for 1.6 and update the design in consequent PR if it is not much work. This feature is hidden behind a setting anyway and it is still Alpha. Wdyth?

evankanderson:
Sorry, I've been wrestling with gateway api and an internal re-org. I was 👍 before the latest changes, but I'll give it another look before approving.

Thanks for doing this! I'm hopeful that we might be able to follow the same approach with the Gateway API fairly easily, since that allows explicitly specifying a listener port.

evankanderson:
I suspect that most tenants will _not_ have access to edit the labels on their namespace; this would also allow bypassing any number of other properties on the namespace. (Similarly, they will not be able to create new namespaces with labels that they want.)

evankanderson:
I'd like to understand what the risk of exposing the listener port assignments is, if NetworkPolicy is set up to disallow access from all but one namespace to that port. The public port numbers of a service are exposed via SRV records which aren't restricted within the cluster:

```bash
# dig srv mysql-db-mysql-master.wordpress.svc.cluster.local @10.96.0.10
...
;; ANSWER SECTION:
mysql-db-mysql-master.wordpress.svc.cluster.local. 30 IN SRV 0 50 3306 mysql-db-mysql-master.wordpress.svc.cluster.local.
mysql-db-mysql-master.wordpress.svc.cluster.local. 30 IN SRV 0 50 8080 mysql-db-mysql-master.wordpress.svc.cluster.local.

;; ADDITIONAL SECTION:
mysql-db-mysql-master.wordpress.svc.cluster.local. 30 IN A 10.101.116.145
```

If an attacker can enumerate the ports (and there are a max of 32k such ports, they can simply probe them all to determine the correct port. On a local network with 1ms RTT, this would take about 90 seconds without parallelism, so it's not much of a defense on its own without NetworkPolicy to block the packets.

pkg/generator/ingress_translator.go
```diff
@@ -58,18 +61,21 @@ type IngressTranslator struct {
 	secretGetter    func(ns, name string) (*corev1.Secret, error)
 	endpointsGetter func(ns, name string) (*corev1.Endpoints, error)
 	serviceGetter   func(ns, name string) (*corev1.Service, error)
+	namespaceGetter func(name string) (*corev1.Namespace, error)
 	tracker         tracker.Interface
 }
```
evankanderson, line 66: Not for this PR, but it feels weird for this to be a struct rather than an interface.

pkg/generator/ingress_translator.go
```diff
@@ -259,12 +265,38 @@ func (translator *IngressTranslator) translateIngress(ctx context.Context, ingre
...
 	return &translatedIngress{
 		name: types.NamespacedName{
 			Namespace: ingress.Namespace,
 			Name:      ingress.Name,
 		},
+		listener:                listener,
+		port:                    port,
 		sniMatches:              sniMatches,
 		clusters:                clusters,
 		externalVirtualHosts:    externalHosts,
```
evankanderson, line 302: Maybe not for this PR, but it feels like this method might work better to construct the return value at the top (with default values for all the calculated fields), and then fill in the struct as it moves through the function. This would avoid needing to (for example) pre-fill `listener`, `port`, `sniMatches`, etc.

pkg/reconciler/ingress/lister.go
```diff
@@ -82,9 +85,24 @@ func (l *gatewayPodTargetLister) getIngressUrls(ing *v1alpha1.Ingress, gatewayIp
 				target.URLs = domainsToURL(domains, scheme)
 			}
 		} else {
+			portPort := strconv.Itoa(int(config.HTTPPortInternal))
```
evankanderson, line 88: Should this be `podPort` rather than `portPort`?

pkg/config/config.go
```diff
@@ -67,6 +67,13 @@ const (
 	// ServingNamespaceEnv is an env variable specifying where the serving is deployed.
 	// e.g. OpenShift deploys Kourier in different namespace so `system.Namespace()` does not work.
 	ServingNamespaceEnv = "SERVING_NAMESPACE"
+
+	// TenantNameLabelKey is the annotation key for grouping/isolating ingresses by tenant
+	TenantNameLabelKey = "admin.knative.dev/tenant-name"
```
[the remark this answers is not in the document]
evankanderson: If this is being used to generate the listener service name, why not use the port number in the service name to avoid needing to correlate a second string (e.g. `kourier-isolation-6452` as a Service name, where the first part is a fixed string)?

evankanderson (commented):
Some minor nits, but I'd be comfortable shipping this in 1.6 to iterate from there.

I'm holding off from a `/lgtm` in case someone else has concerns, or if you want to take my suggestion and name your listening Services based on assigned port numbers to avoid possible `listener` or `tenant-name` collisions.

/approve

nak3:
Hmm... Not only other tenant, but also it is not possible to hijack any Envoy configuration like Envoy's admin interface? I guess that it is impossible with the default static configuration and cannot come up with how exactly we can attack. So, can I trust you?

evankanderson:
It depends on how envoy is configured, but it doesn't seem like this opens _new_ attack surface there.

skonto:
Can we run https://github.com/knative/serving/pull/13094 also with the feature enabled? We need to have an e2e test somewhere, could you add one here at least. Also is there a plan for the tenant controller here? Is there a list of the remaining tasks? In the issue description all tasks seem done. Let's add the tenant controller idea and the webhook validation too pls.
Also since we plan to iterate let's add the question label vs annotation there.
@lionelvillard could you do this before merging?

pkg/config/config.go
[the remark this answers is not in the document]
lionelvillard: This is an excellent point! There is no need for the listener label anymore. `tenant-name` feels wrong in this context anyway so I'm more that happy to remove this label!

lionelvillard:
@skonto 

> Can we run [knative/serving#13094](https://github.com/knative/serving/pull/13094) also with the feature enabled? 

Yes. https://github.com/knative/serving/pull/13112

> We need to have an e2e test somewhere, could you add one here at least.

I'm looking at this now, it seems easy enough but due to the time constraint, I would rather do this in a separate PR.

IMO, as soon as both https://github.com/knative/serving/pull/13094 and https://github.com/knative/serving/pull/13112 are green, it should be safe to merge this PR.

> Also is there a plan for the tenant controller here?

No plans. This is a much bigger topic!

> Let's create a list of the remaining tasks pls (tracking issue). In this description of this issue all tasks seem done. Let's add the tenant controller idea and the webhook validation too. Also since we plan to iterate on this let's add the question about label vs annotation there. @lionelvillard could you do this before merging?

Yes I can create an Github issue tracking all remaining tasks

lionelvillard:
@evankanderson @skonto @nak3 @rhuss Please can you take another look? Since the last round of reviews I:
- removed the `listener` label. The Kourier services exposing the envoy listeners are now called `kourier-isolation-<listener-port>`
- remove "none" in the configmap in favor of the empty string value
- opened various issues to track the remaining tasks

I'm tracking https://github.com/knative/serving/pull/13094 and https://github.com/knative/serving/pull/13112. So far it looks good.

I'm working on e2e tests (in a separate PR). I don't know if I will have time to finish before the 1.6 release (which is tomorrow).

pkg/reconciler/ingress/ingress.go
```diff
@@ -77,6 +80,20 @@ func (r *Reconciler) ReconcileKind(ctx context.Context, ing *v1alpha1.Ingress) r
...
+				if err != nil {
+					return fmt.Errorf("failed to get namespace: %w", err)
+				}
+
+				if ns.Annotations != nil {
+					if port, ok := ns.Annotations[config.ListenerPortAnnotationKey]; ok {
+						internal = config.ListenerServiceHostnames(port)
+					}
+				}
+			}
```
evankanderson, line 95: This could now be extracted into a function since it's shared with `lister.go` 90-110.

(For a future PR)

evankanderson (commented):
A few notes for the next PR, but happy to get this in for 1.6.

/lgtm
/approve

pkg/config/configmap.go
```diff
@@ -90,4 +98,6 @@ type Kourier struct {
 	// this option, for example, the "timeoutSeconds" specified in Knative service is still
 	// valid.
 	IdleTimeout time.Duration
+	// Desire level of incoming traffic isolation
+	TrafficIsolation string
```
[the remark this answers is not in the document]
lionelvillard: it's a typedef now.

pkg/reconciler/ingress/ingress.go
lionelvillard: I'll do this in a future PR. (added to the list of TODOs)

lionelvillard:
@evankanderson can you please TAL? All tests are green (including the Serving ones).

PR #855, opened 2022-06-08
Use `internal-encryption` to deploy internal certificates automatically

nak3:
This patch is same purpose with https://github.com/knative/serving/pull/13005.

Currently users have to deploy certificates manually with several options such as
`activator-san`, `activator-ca`, `queue-proxy-ca` etc.

Such deployment and management of the certificates is a big burden for users.


skonto:
@nak3 pls rebase.

nak3:
@skonto This also needs https://github.com/knative/networking/pull/680.
Could you please review & merge https://github.com/knative/networking/pull/680 first? Otherwise, I need to keep rebasing this PR.

pkg/config/config.go
```diff
@@ -63,6 +63,10 @@ const (
 	// disableHTTP2AnnotationKey is the annotation key attached to a Knative Domain Mapping
 	// to indicate that http2 should not be enabled for it.
 	disableHTTP2AnnotationKey = "kourier.knative.dev/disable-http2"
+
+	// ServingNamespaceEnv is an env variable specifying where the serving is deployed.
+	// e.g. OpenShift deploys Kourier in different namespace so `system.Namespace()` does not work.
```
skonto, line 68: Afaik this holds for backwards compatibility reasons. Is there anything that stops us from moving everything to `serving` ns?

nak3: It is possible but it has a lot of risks to move the gateway into serving ns.

nak3:
@skonto Could you please take a look?

pkg/generator/ingress_translator.go
```diff
@@ -313,7 +310,7 @@ func createUpstreamTLSContext(caCertificate []byte, activatorSAN string, alpnPro
 					},
 					MatchSubjectAltNames: []*envoymatcherv3.StringMatcher{{
 						MatchPattern: &envoymatcherv3.StringMatcher_Exact{
-							Exact: activatorSAN,
+							Exact: certificates.FakeDnsName,
```
skonto, line 313: I guess if we ever have multiple-knative installations this will have to be changed?

PR #859, opened 2022-06-10
bump kind and k8s versions in kind-e2e tests

psschwei:
Signed-off-by: Paul S. Schweigert <paulschw@us.ibm.com>

Since we will probably be supporting Kubernetes v1.24 in the next release, and kind v0.13.0+ is required to run Kubernetes v1.24.0+ images, let's go ahead and bump the versions now, so we have a chance to work out any issues before the next release.


carlisia:
👍
/lgtm
/approve

PR #862, opened 2022-06-23
Support secret filtering informer

skonto:
# Changes

- Similar to https://github.com/knative-sandbox/net-istio/pull/920
- Tested via the internal encrypted traffic feature.

Fixes #

**Release Note**

```release-note
Secret informer in net-kourier ingress controller can filter secrets based on the existence of a label key`networking.internal.knative.dev/certificate-uid`. Users are required to add the label key for custom secrets defined in ingresses.
For special cases users are recommended to reach out to Knative community.
In case of AutoTLS provided by Knative net-certmanager or net-http01 this is done transparently.
To enable this behavior at net-kourier set env var ENABLE_SECRET_INFORMER_FILTERING_BY_CERT_UI to true. 
This will be available by default in future releases.
```

**Docs**

```docs
Secret informer in net-kourier ingress controller can filter secrets based on the existence of a label key`networking.internal.knative.dev/certificate-uid"`. Users are required to add the label key for custom secrets defined in ingresses.
For special cases users are recommended to reach out to Knative community.
In case of AutoTLS provided by Knative net-certmanager or net-http01 this is done transparently.
To enable this behavior at net-kourier set env var ENABLE_SECRET_INFORMER_FILTERING_BY_CERT_UI to true. 
This will be available by default in future releases.
```


skonto:
Tested in: https://github.com/knative/serving/pull/13058

nak3:
@skonto Could you update the `Release Note` and `Docs` by `s/net-istio/net-kourier`?

pkg/reconciler/ingress/controller.go
```diff
@@ -42,8 +43,10 @@ import (
 	kubeclient "knative.dev/pkg/client/injection/kube/client"
 	endpointsinformer "knative.dev/pkg/client/injection/kube/informers/core/v1/endpoints"
 	podinformer "knative.dev/pkg/client/injection/kube/informers/core/v1/pod"
-	secretinformer "knative.dev/pkg/client/injection/kube/informers/core/v1/secret"
+	secretfilteredinformer "knative.dev/pkg/client/injection/kube/informers/core/v1/secret/filtered"
 	serviceinformer "knative.dev/pkg/client/injection/kube/informers/core/v1/service"
+	filteredFactory "knative.dev/pkg/client/injection/kube/informers/factory/filtered"
+
```
nak3, line 49: Can you remove this blank line?

test/generate-upstream-cert.sh
```diff
@@ -33,8 +33,10 @@ openssl x509 -req -extfile <(printf "subjectAltName=DNS:$san") -days 365 -in "${
...
 kubectl create -n ${SERVING_SYSTEM_NAMESPACE} secret generic knative-serving-certs \
-    --from-file=ca-cert.pem="${out_dir}"/root.crt --dry-run=client -o yaml | kubectl apply -f -
+    --from-file=ca-cert.pem="${out_dir}"/root.crt --dry-run=client -o yaml |  \
+    sed  '/^metadata:/a\ \ labels: {"networking.internal.knative.dev/certificate-uid":"test-id"}' | kubectl apply -f -
 
 kubectl create -n ${TEST_NAMESPACE} secret tls server-certs \
     --key="${out_dir}"/tls.key \
-    --cert="${out_dir}"/tls.crt --dry-run=client -o yaml | kubectl apply -f -
+    --cert="${out_dir}"/tls.crt --dry-run=client -o yaml | \
+    sed  '/^metadata:/a\ \ labels: {"networking.internal.knative.dev/certificate-uid":"test-id"}' | kubectl apply -f -
```
nak3, line 42: I think we don't need to add the label to the `server-certs`. It is used by users' namespace and Kourier does not need to be reconciled when they updated them.

(If users started using a server cert signed by a different CA, Kourier needs to be reconciled to load the new CA but it is triggered by `knative-serving-certs`(=CA) change.)

nak3: With said, the test will fail without this label? :sweat_smile:

skonto: Kourier reconciles a related ingress referencing that secret through no? Afaik it does fail I can try once more without it.

nak3: Yes for the CA secret (=`knative-serving-certs`) , but should be no for the server cert (=`server-certs`).
But the test failed? Hmm... I misunderstood something. I will look into it as well.

skonto: Let me try it first. I might remember wrongly.

skonto: Ok removed the label and switched on filtering.

skonto: Tests passed reverting the default value for the env var.

PR #866, opened 2022-07-04
Add timeout config to Kourier

### LINKED ISSUE #680: Kourier is not compatible with Knative-Service timeoutSeconds custom value with more than 300 seconds
When a long GRPC connection is established, the data does not return within 5 minutes, and the connection will be interrupted, which can be solved by adjusting the timeoutSeconds parameter. However, when kourier is used as a gateway, the timeoutSeconds value is fixed to 300 seconds by default and cannot be customized.

In use, it is found that in kourier v0.26.0 version, kourier does not support the custom parameter of timeoutSeconds in Knative-Service.

Also does not support, the max-revision-timeout-seconds and revision-timeout-seconds custom parameters in knative-serving comfigmap config-defaults.

```
apiVersion: serving.knative.dev/v1
kind: Service
metadata:
  name: test
spec:
  template:
    metadata:
      annotations:
        autoscaling.knative.dev/class: kpa.autoscaling.knative.dev
        autoscaling.knative.dev/metric: concurrency
        autoscaling.knative.dev/minScale: "0"
        autoscaling.knative.dev/maxScale: "1"
    spec: 
      timeoutSeconds: 600   # kourier gw is not compatible with this custom number.
      containers:
      - image: imageurl
        imagePullPolicy: IfNotPresent
        ports:
          - name: h2c
            containerPort: 8080
```

**Reminder:** Istio gateway is compatible with timeoutSeconds custom value.

xxinran:
Fix #680

### DIFF
--- a/config/200-config.yaml
+++ b/config/200-config.yaml
@@ -60,4 +60,4 @@ data:
     cluster-cert-secret: ""
 
     # Specifies the amount of time that Kourier waits for the incoming requests.
-    stream-idle-timeout: 300s
+    stream-idle-timeout: "300s"
--- a/config/200-config.yaml
+++ b/config/200-config.yaml
@@ -58,3 +58,6 @@ data:
     # NOTE: This flag is in an alpha state and is mostly here to enable internal testing
     #       for now. Use with caution.
     cluster-cert-secret: ""
+
+    # Specifies the amount of time that Kourier waits for the incoming requests.
+    stream-idle-timeout: 300s
--- a/pkg/config/configmap.go
+++ b/pkg/config/configmap.go
@@ -17,6 +17,8 @@ limitations under the License.
 package config
 
 import (
+	"time"
+
 	corev1 "k8s.io/api/core/v1"
 
 	cm "knative.dev/pkg/configmap"
@@ -35,6 +37,10 @@ const (
 
 	// clusterCert is the config map key for kourier internal certificates
 	clusterCert = "cluster-cert-secret"
+
+	// IdleTimeoutKey is the config map key for the amount of time that Kourier waits
+	// for incoming requests. This value is set to "stream_idle_timeout" in Envoy.
+	IdleTimeoutKey = "stream-idle-timeout"
 )
 
 func DefaultConfig() *Kourier {
@@ -42,6 +48,7 @@ func DefaultConfig() *Kourier {
 		EnableServiceAccessLogging: true, // true is the default for backwards-compat
 		EnableProxyProtocol:        false,
 		ClusterCertSecret:          "",
+		IdleTimeout:                300 * time.Second, // default value
 	}
 }
 
@@ -53,6 +60,7 @@ func NewConfigFromMap(configMap map[string]string) (*Kourier, error) {
 		cm.AsBool(enableServiceAccessLoggingKey, &nc.EnableServiceAccessLogging),
 		cm.AsBool(enableProxyProtocol, &nc.EnableProxyProtocol),
 		cm.AsString(clusterCert, &nc.ClusterCertSecret),
+		cm.AsDuration(IdleTimeoutKey, &nc.IdleTimeout),
 	); err != nil {
 		return nil, err
 	}
@@ -76,4 +84,10 @@ type Kourier struct {
 	// ClusterCertSecret specifies the secret name for the server certificates of
 	// Kourier Internal.
 	ClusterCertSecret string
+	// IdleTimeout specifies the amount of time that Kourier waits for incoming requests.
+	// The default value is 5 minutes. This will not interfere any smaller configured
+	// timeouts that may have existed in configurations prior to
+	// this option, for example, the "timeoutSeconds" specified in Knative service is still
+	// valid.
+	IdleTimeout time.Duration
 }
--- a/pkg/envoy/api/http_connection_manager.go
+++ b/pkg/envoy/api/http_connection_manager.go
@@ -35,7 +35,7 @@ import (
 
 // NewHTTPConnectionManager creates a new HttpConnectionManager that points to the given
 // RouteConfig for further configuration.
-func NewHTTPConnectionManager(routeConfigName string, enableAccessLog, enableProxyProtocol bool) *hcm.HttpConnectionManager {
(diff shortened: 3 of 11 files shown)


nak3 (changes requested):
Thank you so much!

Could you please change this PR to use the timeout in Kourier?

config/200-config.yaml
```diff
@@ -58,3 +58,6 @@ data:
     # NOTE: This flag is in an alpha state and is mostly here to enable internal testing
     #       for now. Use with caution.
     cluster-cert-secret: ""
+
+    # Specifies the amount of time that Kourier waits for the incoming requests.
+    stream_idle_timeout: 300
```
skonto, line 63: a)  We need to document this and the end-to-end timeout settings in general.
For example I assume that users will need to adapt this setting if they also configure revision time out. 
b) There is also `max_stream_duration` from what I read, do we need to set that to the `DefaultMaxRevisionTimeoutSeconds`.

xxinran: Thanks for your comments @skonto
a) Sure, I will add a detailed description of this option. As for `end-to-end timeout`, I guess you mean the `timeoutSeconds` in knative service, now we could simply set it to "0" to disable the timeout check at ingress level. In the future,  it is possible to overwrite `stream_idile_timeout` with `timeoutSecond` if it is longer than `timeoutSecond`
b) As for other timeout configs, I think we could consider them later and implement them in other PRs if needed.

pkg/config/configmap.go
```diff
@@ -35,13 +37,17 @@ const (
...
+	// timeOut is the amount of time that Kourier waits for incoming requests.
+	IdleTimeoutKey = "stream_idle_timeout"
 )
 
 func DefaultConfig() *Kourier {
 	return &Kourier{
 		EnableServiceAccessLogging: true, // true is the default for backwards-compat
 		EnableProxyProtocol:        false,
 		ClusterCertSecret:          "",
+		IdleTimeout:                0, // infinity
```
nak3, line 50: The current idle timeout is 300, isn't it? If so, we should default to `300` in this PR.

We should carefully change the default to infinity later in another PR as it would affect current users.

pkg/envoy/api/http_connection_manager.go
```diff
@@ -35,7 +35,7 @@ import (
 
 // NewHTTPConnectionManager creates a new HttpConnectionManager that points to the given
 // RouteConfig for further configuration.
-func NewHTTPConnectionManager(routeConfigName string, enableAccessLog, enableProxyProtocol bool) *hcm.HttpConnectionManager {
+func NewHTTPConnectionManager(routeConfigName string, enableAccessLog, enableProxyProtocol bool, idleTimeout time.Duration) *hcm.HttpConnectionManager {
```
nak3, line 38: Could you change the arguments to take `config.Kourier` instead of `enableAccessLog, enableProxyProtocol bool, idleTimeout time.Duration`?

config/200-config.yaml
```diff
@@ -58,3 +58,6 @@ data:
     # NOTE: This flag is in an alpha state and is mostly here to enable internal testing
     #       for now. Use with caution.
     cluster-cert-secret: ""
+
+    # Specifies the amount of time that Kourier waits for the incoming requests.
+    stream_idle_timeout: 300s
```
nak3, line 63: Snake case is used by Envoy but Knative usually use kebab case for the configuration. The configuration should be consistent.

config/200-config.yaml
```diff
@@ -58,3 +58,6 @@ data:
     # NOTE: This flag is in an alpha state and is mostly here to enable internal testing
     #       for now. Use with caution.
     cluster-cert-secret: ""
+
+    # Specifies the amount of time that Kourier waits for the incoming requests.
```
nak3, line 62: I think that we should mention that the specified value is set to `stream_idle_timeout` in Envoy, though it is obvious from the configuration key.

nak3 (commented):
Thank you! Could you add a simple e2e test with the timeout in https://github.com/knative-sandbox/net-kourier/blob/main/test/e2e-kind.sh ?

xxinran:
> Thank you! Could you add a simple e2e test with the timeout in https://github.com/knative-sandbox/net-kourier/blob/main/test/e2e-kind.sh ?

Sure, I prefer to add tests in a separate PR later. how about this?

.vscode/settings.json
```diff
@@ -0,0 +1,3 @@
+{
```
nak3, line 1: Could you remove this?

nak3 (commented):
The code looks good. But I haven't tested this actually. It is easily verified if we have the e2e test, though...

test/e2e-kind.sh
```diff
@@ -133,3 +133,13 @@ go test -race -count=1 -timeout=5m -tags=e2e ./test/proxyprotocol/... \
   --ingressendpoint="${IPS[0]}" \
   --ingressClass=kourier.ingress.networking.knative.dev \
   --cluster-suffix="$CLUSTER_SUFFIX"
+
+echo ">> Set IdleTimeout to 50s"
+kubectl -n "${KOURIER_CONTROL_NAMESPACE}" patch configmap/config-kourier --type merge -p '{"data":{"stream-idle-timeout":"50s"}}'
+kubectl -n "${KOURIER_CONTROL_NAMESPACE}" rollout restart deployment/net-kourier-controller
+
+echo ">> Running IdleTimeout tests"
+go test -v  -tags=e2e ./test/e2e -run "^TestIdleTimeout*" \
```
nak3, line 142: Could you move the test into `test/e2e/timeout/` and run the test as `go test -v -tags=e2e ./test/e2e/timeout` ?

test/e2e-kind.sh
```diff
@@ -133,3 +133,14 @@ go test -race -count=1 -timeout=5m -tags=e2e ./test/proxyprotocol/... \
   --ingressendpoint="${IPS[0]}" \
   --ingressClass=kourier.ingress.networking.knative.dev \
   --cluster-suffix="$CLUSTER_SUFFIX"
+
```
nak3, line 136: Let's revert `enable-proxy-protocol` to `false` and run the test.

nak3:
/lgtm
/approve

Thank you for this PR and your patience!

PR #869, opened 2022-07-05
Proxy protocol and SNI do not work together

### LINKED ISSUE #868: SNI does not work when proxy_protocol is enabled
SNI and proxy_protocol features do not work together and we are getting the wrong certificate.
The reason behind that is that the TLS inspector listener filter is executed before proxy_protocol listener feature:
```
listener_filters:
  - name: envoy.filters.listener.tls_inspector
  - name: envoy.filters.listener.proxy_protocol
    typed_config:
         '@type': >-
         type.googleapis.com/envoy.extensions.filters.listener.proxy_protocol.v3.ProxyProtocol
```

# How To Reproduce The Problem
```
- enable auto TLS with cert-manager
- enable proxy-protocol feature
- create a ksvc
- create a custom domain (domain mapping)
- invoke your ksvc using Domain Mapping URL
```

# How To fix The Problem
proxy_protocol listener filter should be executed before TLS inspector listener filter
```
listener_filters:
  - name: envoy.filters.listener.proxy_protocol
    typed_config:
         '@type': >-
         type.googleapis.com/envoy.extensions.filters.listener.proxy_protocol.v3.ProxyProtocol
  - name: envoy.filters.listener.tls_inspector
```

RamyChaabane:
# Changes

- :bug: Fix bug
Proxy protocol and SNI do not work together, because the TLS inspector listener filter is executed before proxy_protocol listener filter

/kind bug

Fixes #868

**Release Note**

```release-note
fix SNI with proxy_protocol
```

### DIFF
--- a/test/proxyprotocol/proxyprotocol_test.go
+++ b/test/proxyprotocol/proxyprotocol_test.go
@@ -33,6 +33,7 @@ import (
 	"gotest.tools/v3/assert"
 	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	ktypes "k8s.io/apimachinery/pkg/types"
 	"k8s.io/apimachinery/pkg/util/intstr"
 	"k8s.io/apimachinery/pkg/util/wait"
 	"knative.dev/networking/pkg/apis/networking"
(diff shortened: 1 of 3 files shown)


nak3:
Thank you so much!

@RamyChaabane Do you think It is possible to add and e2e test for this in https://github.com/knative-sandbox/net-kourier/tree/main/test/proxyprotocol ?

RamyChaabane:
Hello @nak3 
I've tried to add an e2e test but it doesn’t work and I don't know what I'm doing wrong :/

test/proxyprotocol/proxyprotocol_test.go
```diff
@@ -91,6 +91,77 @@ func TestProxyProtocol(t *testing.T) {
 	assert.Equal(t, resp.StatusCode, http.StatusOK)
 }
 
+// TestProxyProtocolWithSNI verifies that the kourier is configured with proxy protocol and SNI.
+func TestProxyProtocolWithSNI(t *testing.T) {
+	ctx, clients := context.Background(), test.Setup(t)
+	privateHostname := test.ObjectNameForTest(t) + "." + test.ServingNamespace + ".svc." + test.NetworkingFlags.ClusterSuffix
+	name, port, _ := ingress.CreateProxyService(ctx, t, clients, privateHostname, "kourier-internal.kourier-system.svc.cluster.local")
```
nak3, line 98:
```suggestion
        name, port, _ := ingress.CreateProxyService(ctx, t, clients, privateHostname, "kourier-internal.kourier-system.svc."+test.NetworkingFlags.ClusterSuffix)
```

kind test does not use `cluster.local` so we should not use `clsuter.local`. https://github.com/knative-sandbox/net-kourier/blob/main/.github/workflows/kind-e2e.yaml#L55-L57

nak3, line 98: `kourier-internal` should be replaced with `InternalServiceName` defined in `pkg/config/config.go`.

`kourier-system` also should be replaced with constant value. `test/ha/kourier_gateway_test.go` already has the value so we should move the const value somewhere in common place?

RamyChaabane:
Hello @nak3 
better but I got 404 status error. So I do it differently and now it works

nak3:
/lgtm
/approve

Thank you!

PR #878, opened 2022-07-15
Set no limit to the timeout by default

### LINKED ISSUE #874: Set no limit to the timeout by default
Follow up for https://github.com/knative-sandbox/net-kourier/issues/680 and https://github.com/knative-sandbox/net-kourier/pull/866

The timeout can be configured via `stream-idle-timeout` in config-kourier but the default value is still `300s` now.
So users have to configure the timeout value more than 300s when they set more than 300s `timeoutSeconds` in Knative Serving.

We probably should set the default to infinity if there are no any concerns..

_NOTE:_ The timeout of Contour and Istio are infinity by default.

xxinran:
Fixes #874 

# Changes
Change the default value of timeout to "infinity".

**Release Note**
```release-note
The timeout can be configured via stream-idle-timeout in config-kourier, the default value is set to "infinity" which means there is no timeout. 
Users can change this value in config-kourier with decimal numbers plus a unit suffix such as "ns", "us" (or "µs"), "ms", "s", "m", "h"
```

### DIFF
--- a/config/200-config.yaml
+++ b/config/200-config.yaml
@@ -60,7 +60,8 @@ data:
     cluster-cert-secret: ""
 
     # Specifies the amount of time that Kourier waits for the incoming requests.
-    stream-idle-timeout: "infinity"
+    # The default, 0s, imposes no timeout at all.
+    stream-idle-timeout: "0s"
 
     # Control the desired level of incoming traffic isolation.
     #
--- a/config/200-config.yaml
+++ b/config/200-config.yaml
@@ -60,7 +60,7 @@ data:
     cluster-cert-secret: ""
 
     # Specifies the amount of time that Kourier waits for the incoming requests.
-    stream-idle-timeout: "300s"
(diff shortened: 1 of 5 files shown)


config/200-config.yaml
```diff
@@ -60,7 +60,7 @@ data:
     cluster-cert-secret: ""
 
     # Specifies the amount of time that Kourier waits for the incoming requests.
-    stream-idle-timeout: "300s"
+    stream-idle-timeout: "infinity"
```
nak3, line 63:
```suggestion
    # The default, 0s, imposes no timeout at all.
    stream-idle-timeout: "0s"
```

I think we can just put a comment about `0s` without `asKourierDuration()` which is simple, isn't it?

nak3 (commented):
Thank you so much!
Commented about one tiny thing but LGTM.

nak3:
/lgtm
/approve

Thank you!

Issue #840, opened 2022-05-10
net-kourier-controller component liveliness probe Failing


junydania:
Having an issue with net-kourier-knative liveliness probe failing.  Describe the pod gives this data

```
  Normal   Started    65s   kubelet            Started container controller
  Warning  Unhealthy  64s   kubelet            Readiness probe failed: 2022/05/10 14:25:56 failed to connect to service at ":18000": context deadline exceeded
  Warning  Unhealthy  56s   kubelet            Readiness probe failed: 2022/05/10 14:26:04 failed to connect to service at ":18000": context deadline exceeded
  Warning  Unhealthy  46s   kubelet            Readiness probe failed: 2022/05/10 14:26:14 failed to connect to service at ":18000": context deadline exceeded
  Warning  Unhealthy  36s   kubelet            Readiness probe failed: 2022/05/10 14:26:24 failed to connect to service at ":18000": context deadline exceeded
  Warning  Unhealthy  26s   kubelet            Readiness probe failed: 2022/05/10 14:26:34 failed to connect to service at ":18000": context deadline exceeded
  Warning  Unhealthy  16s   kubelet            Readiness probe failed: 2022/05/10 14:26:44 failed to connect to service at ":18000": context deadline exceeded
  Warning  Unhealthy  6s    kubelet            Readiness probe failed: 2022/05/10 14:26:54 failed to connect to service at ":18000": context deadline exceeded
```

logs from the pod

```
{"severity":"INFO","timestamp":"2022-05-10T14:25:56.629935147Z","caller":"logging/config.go:116","message":"Successfully created the logger."}
{"severity":"INFO","timestamp":"2022-05-10T14:25:56.629991527Z","caller":"logging/config.go:117","message":"Logging level set to: info"}
{"severity":"INFO","timestamp":"2022-05-10T14:25:56.630065312Z","caller":"logging/config.go:79","message":"Fetch GitHub commit ID from kodata failed","error":"open /var/run/ko/HEAD: no such file or directory"}
{"severity":"INFO","timestamp":"2022-05-10T14:25:56.630091394Z","logger":"net-kourier-controller","caller":"profiling/server.go:64","message":"Profiling enabled: false"}
{"severity":"INFO","timestamp":"2022-05-10T14:25:56.634867894Z","logger":"net-kourier-controller","caller":"leaderelection/context.go:46","message":"Running with Standard leader election"}
[...]

PR #879, opened 2022-07-18
make ingressSync run in parallel

junydania:
# Changes

Executes the ingressSync in parallel. Suggesting this change due to a problem we encountered with net-kourier priming ingresses.

With over 1,500 services in a cluster, it takes over 20-30mins for this to complete. 

Though we can run this process in parallel


nak3:
Thank you so much @junydania Did this PR solve https://github.com/knative-sandbox/net-kourier/issues/840 ?
How did you measure that this PR improved the performance?

PR #889, opened 2022-08-03
Add trusted hops parameters support

thomas-tacquet:
# Changes

- :gift: Add support for trusted hops parameter

/kind enhancement

**Release Note**

```release-note
- add support for trusted hops count parameter
```

**Docs**

```docs
This alow users to have trusted hops support
```


thomas-tacquet:
@nak3 can you check this please ? 🙏

nak3:
Thank you! I will take a look after back from the holiday on this Thursday.

nak3 (commented):
Do you think there is any way to add an e2e test for this? I think we cannot add it simply but please let me double check with you.

thomas-tacquet:
Hello @nak3  🧇 

About testing trusted hops support, in my environment on my cluster I validated trusted hops (0, 1, 2+...) using different layers of proxies to ensure the parameter is read and appliyed.

But in e2e tests I did not succed building such a way of testing 😬 I tried to find how other project implemented trusted hops testing but I did not find anything relevant that could help me.

Units tests in this PR ensure the parameter is read but it does not guarantee everyone that trusted hops works as expected 🤔 

I think we can merge this because I'm confident with this parameter adding and adding e2e tests for this can be done later ?

thomas-tacquet:
Hello @nak3 👋 

Is this PR acceptable with current testing coverage ? Unit testing is ok but I can't figure out how to properly integrade good E2E tests for this case 🤔 

(IMHO ensuring the parameter is correctly set is "enough" as this PR scope is to add the parameter and not to implement the trusted hops logic.)

nak3:
Sorry for the delay.
The PR looks good to me but I think we can add one tiny unit test in https://github.com/knative-sandbox/net-kourier/blob/main/pkg/envoy/api/http_connection_manager_test.go ?

config/200-config.yaml
```diff
@@ -80,3 +80,7 @@ data:
     # NOTE THAT THIS IS AN EXPERIMENTAL / ALPHA FEATURE.
     enable-cryptomb: "false"
 
+
+    # trusted-hops-count Configure the number of additional ingress proxy hops from the
```
nak3, line 84:
```suggestion
    # Configures the number of additional ingress proxy hops from the
```

We should align the format with the other values.

pkg/config/configmap.go
```diff
@@ -108,6 +110,9 @@ type Kourier struct {
 	IdleTimeout time.Duration
 	// Desire level of incoming traffic isolation
 	TrafficIsolation TrafficIsolationType
+	// TrustedHopsCount Configure the number of additional ingress proxy hops from the
```
nak3, line 113:
```suggestion
	// TrustedHopsCount configures the number of additional ingress proxy hops from the
```

nit

thomas-tacquet:
No problem @nak3 😁 
I added an unit test to check for different values (0, 1 and max 👌)

pkg/envoy/api/http_connection_manager_test.go
```diff
@@ -110,3 +111,40 @@ func TestNewRouteConfig(t *testing.T) {
 
 	assert.DeepEqual(t, got, want, protocmp.Transform())
 }
+
+func TestNewHTTPConnectionManagerWithoutTrustedHopsDisabled(t *testing.T) {
```
nak3, line 115:
```suggestion
func TestNewHTTPConnectionManagerWithTrustedHops(t *testing.T) {
```
Your test name is also fine so please feel free to apply the suggestion or not by your preference.

nak3:
/lgtm
/approve

(I think I don't have the approve permission now...)

pkg/envoy/api/http_connection_manager_test.go
thomas-tacquet: Nice spot, at first I wanted to create different test and then moved to a single loop test 😁

evankanderson (commented):
/lgtm
/approve

If you wanted to write an integration test, I would simply have a client send a request with 2-3 `x-forwarded-for` levels of headers, and have the receiving program echo back the resulting headers in the request body.  The client could then determine if the correct set of headers had been accepted.

In any case, this seems reasonable, so approving.

evankanderson:
@nak3 -- you should still be able to have approvers if you want it, IMO.  (Though we should check with @dprotaso and @psschwei )

Let me know if you want that bit -- we can add you either to `networking-approvers` or specifically for `net-kourier`.

nak3:
Thank you! Yes, I noticed that it was just dropped (by Dave's comment) and added myself to the approver yesterday https://github.com/knative/community/pull/1249

PR #898, opened 2022-08-17
deliberately drop a path or multiple paths

### LINKED ISSUE #897: Should be able to drop a path in the ingress level
Hello there,

Today you can drop a path in your application by implementing a route that returns 404 http code status when invoking your Knative service on that path. If you have configured `scale to zero` feature, this behaviour is not optimal as your pod will scale and you will be waiting for a few seconds (cold start time + time till your pod is in runnig state) for your path to be dropped.

I think we should be doing that in the ingress level

RamyChaabane:
# Changes

:gift: Add new feature

/kind enhancement

Fixes #897

**Release Note**

```release-note
deliberately drop a/multiple path(s)
action required: user will need to add a new annotation to his KSVC to include the path(s) that need to be dropped
```

**Docs**

```docs
Documentation on how configure this is included in PR
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -216,13 +216,22 @@ ingress namespace:
 - `kourier.knative.dev/listener-port`: the envoy listener port
 
 ## Tips
-Domain Mapping is configured to explicitly use `http2` protocol only. This behaviour can be disabled by adding the following annotation to the Domain Mapping resource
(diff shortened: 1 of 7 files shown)


nak3:
Let's continue to discuss this in https://github.com/knative/serving/issues/13228

Issue #274, opened 2020-10-28
Improve security constraints of the envoy image


markusthoemmes:
It currently needs to run as root and needs write access to the file system. Can we improve that?

See https://github.com/knative-sandbox/net-kourier/pull/272 for reference.

nak3:
/close
This was done by https://github.com/knative-sandbox/net-kourier/pull/934 on upstream.

PR #902, opened 2022-08-23
Set runAsNonRoot to true on Kourier Gateway

nak3:
This patch sets runAsNonRoot to false on Kourier Gateway.

As per https://github.com/knative-sandbox/net-kourier/pull/272, the setting was necessary as:
- the envoy image wants to run as root.
- the envoy needs write access to the filesystem.

For write access, I could produce the error if I changed readOnlyRootFilesystem to `false`.

```
[2022-08-22 07:08:29.879][1][critical][main] [external/envoy/source/server/server.cc:108] error initializing configuration '/tmp/config/envoy-bootstrap.yaml': cannot bind '/tmp/envoy.admin': Read-only file system
[2022-08-22 07:08:29.879][1][info][main] [external/envoy/source/server/server.cc:790] exiting
```

However, `runAsNonRoot` could be false without any isssue on OpenShift.
https://github.com/openshift-knative/serverless-operator/pull/1677 verified.

Fix https://github.com/knative-sandbox/net-kourier/issues/274

### DIFF
--- a/config/300-gateway.yaml
+++ b/config/300-gateway.yaml
@@ -68,7 +68,7 @@ spec:
           securityContext:
             allowPrivilegeEscalation: false
             readOnlyRootFilesystem: false
-            runAsNonRoot: false
+            runAsNonRoot: true
             capabilities:
               drop:
                 - all


nak3:
/hold

I think we should merge at the release day. (Today is the release day.)

nak3:
Ah, I see. `runAsNonRoot: true` could work on OpenShift but does not work with upstream envoy image.

```
  Normal   Scheduled  4m6s                  default-scheduler  Successfully assigned kourier-system/3scale-kourier-gateway-8499649b5b-slm86 to kind-worker
  Warning  Failed     118s (x12 over 4m6s)  kubelet            Error: container has runAsNonRoot and image will run as root (pod: "3scale-kourier-gateway-8499649b5b-slm86_kourier-system(d3c4788b-5e98-4676-a04d-467d86413a1a)", container: kourier-gateway)
  Normal   Pulled     104s (x13 over 4m6s)  kubelet            Container image "docker.io/envoyproxy/envoy:v1.20-latest" already present on machine
```

skonto:
You could use a specific user I guess as in https://github.com/projectcontour/contour/pull/4084?

nak3:
No, because of OpenShift's feature.

PR #905, opened 2022-08-24
Made TLS 1.2 the minimum protocol version

sidharthramesh:
# Changes
- Set TLS v1.2 as the default minimal version 

/kind enhancement

As per https://github.com/knative-sandbox/net-kourier/pull/442#issuecomment-790291021, TLS 1.2 should be the default version. envoyproxy has adopted this change in https://github.com/envoyproxy/envoy/commit/f8baa480ec9c6cbaa7a9d5433102efb04145cfc8 (v1.23) - however, Kourier is still using [v1.20](https://github.com/knative-sandbox/net-kourier/blob/09b107be00adb153ef413e6de84c8b3f39b2c9b6/config/300-gateway.yaml#L50).

Fixes #

**Release Note**

```release-note
set TLS v1.2 as the default minimal version 
```

### DIFF
--- a/pkg/envoy/api/listener.go
+++ b/pkg/envoy/api/listener.go
@@ -217,6 +217,9 @@ func createTLSContext(certificate []byte, privateKey []byte) *auth.DownstreamTls
 	return &auth.DownstreamTlsContext{
 		CommonTlsContext: &auth.CommonTlsContext{
 			AlpnProtocols: []string{"h2", "http/1.1"},
+			TlsParams: &auth.TlsParameters{
+				TlsMinimumProtocolVersion: auth.TlsParameters_TLSv1_2,
+			},
 			TlsCertificates: []*auth.TlsCertificate{{
 				CertificateChain: &core.DataSource{
 					Specifier: &core.DataSource_InlineBytes{InlineBytes: certificate},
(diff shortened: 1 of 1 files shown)


pkg/envoy/api/listener.go
```diff
@@ -217,6 +217,9 @@ func createTLSContext(certificate []byte, privateKey []byte) *auth.DownstreamTls
 	return &auth.DownstreamTlsContext{
 		CommonTlsContext: &auth.CommonTlsContext{
 			AlpnProtocols: []string{"h2", "http/1.1"},
+			TlsParams: &auth.TlsParameters{
+				TlsMinimumProtocolVersion: auth.TlsParameters_TLSv1_2,
```
nak3, line 221: Could you add a comment that we don't need this once envoy image is newer than 1.23 as you commented in the description?
> "envoyproxy has adopted this change in https://github.com/envoyproxy/envoy/commit/f8baa480ec9c6cbaa7a9d5433102efb04145cfc8 (v1.23)"

sidharthramesh: Done

nak3:
/lgtm
/approve
Thank you so much!
I opened https://github.com/knative-sandbox/net-kourier/issues/907 as well. If you would like to tackle it, please feel free to take it.

PR #917, opened 2022-09-16
Drop namespace from net-kourier ClusterRole

### LINKED ISSUE #916: ClusterRole should not be namespaced
Kubernetes completely ignores the "namespace" property for cluster-wide resources such as ClusterRole.

This is not an issue when using `kubectl apply`, however some tools like the [Kubernetes Provider for Terraform](https://github.com/hashicorp/terraform-provider-kubernetes#kubernetes-provider-for-terraform--) will do an extra validation step enforcing that cluster-wide resources are not namespaced, and applying the manifest will fail.

Please consider dropping the following line: https://github.com/knative-sandbox/net-kourier/blob/ad9c11ed73d5065a23c361a183490992bc110551/config/200-serviceaccount.yaml#L30

Hellzed:
# Changes
- :broom: Drop namespace from net-kourier ClusterRole

/kind cleanup

Fixes #916 

**Release Note**

```release-note
Drop namespace from net-kourier ClusterRole for better compatibility with Kubernetes Provider for Terraform
```

### DIFF
--- a/config/200-serviceaccount.yaml
+++ b/config/200-serviceaccount.yaml
@@ -27,7 +27,6 @@ apiVersion: rbac.authorization.k8s.io/v1
 kind: ClusterRole
 metadata:
   name: net-kourier
-  namespace: knative-serving
   labels:
     networking.knative.dev/ingress-provider: kourier
     app.kubernetes.io/component: net-kourier


nak3:
Thank you so much @Hellzed !
It seems you need to sign in EasyCLA.

PR #920, opened 2022-09-22
Add keepalive config in bootstrap.yaml

nak3:
As per title, this patch adds the keepalive configuration recommended by
https://www.envoyproxy.io/docs/envoy/latest/api-docs/xds_protocol. 

**Release Note**

```release-note
keepalive config for xDS is added in `kourier-bootstrap` by default. 
```

### DIFF
--- a/config/200-bootstrap.yaml
+++ b/config/200-bootstrap.yaml
@@ -83,6 +83,16 @@ data:
                     pipe:
                       path: /tmp/envoy.admin
         - name: xds_cluster
+          # This keepalive is recommended by envoy docs.
+          # https://www.envoyproxy.io/docs/envoy/latest/api-docs/xds_protocol
+          typed_extension_protocol_options:
+            envoy.extensions.upstreams.http.v3.HttpProtocolOptions:
+              "@type": type.googleapis.com/envoy.extensions.upstreams.http.v3.HttpProtocolOptions
+              explicit_http_config:
+                http2_protocol_options:
+                  connection_keepalive:
+                    interval: 30s
+                    timeout: 5s
           connect_timeout: 1s
           type: strict_dns
           load_assignment:


nak3:
For the context, downstream is seeing sporadic test failures (various tests, but similar symptoms), with ksvcs not becoming Ready due to `Reason:Uninitialized Message:Waiting for load balancer to be ready`. And config dumps showed that only one of the gateway pod had the updated config.

config/200-bootstrap.yaml
```diff
@@ -83,6 +83,16 @@ data:
                     pipe:
                       path: /tmp/envoy.admin
         - name: xds_cluster
+          # This keepalive is recommended by envoy docs.
+          # https://www.envoyproxy.io/docs/envoy/latest/api-docs/xds_protocol
+          typed_extension_protocol_options:
+            envoy.extensions.upstreams.http.v3.HttpProtocolOptions:
+              "@type": type.googleapis.com/envoy.extensions.upstreams.http.v3.HttpProtocolOptions
+              explicit_http_config:
+                http2_protocol_options:
```
maschmid, line 92: I wonder, there is already a `http2_protocol_options` key in the yaml below, would it clash with this one

config/200-bootstrap.yaml
```diff
@@ -83,6 +83,16 @@ data:
...
+                  connection_keepalive:
+                    interval: 30s
+                    timeout: 5s
           connect_timeout: 1s
           type: strict_dns
```
maschmid, line 97: just noticed, there are two `type` keys here, one with `strict_dns` and one with `STRICT_DNS` below

config/200-bootstrap.yaml
nak3: The `http2_protocol_options` below is deprecated one as per https://github.com/envoyproxy/envoy/issues/15559 and https://github.com/envoyproxy/envoy/pull/15563 updated the doc.

I will remove the deprecated one later. (I will send another PR to clean up with `strict_dns` vs `STRICT_DNS`.)

config/200-bootstrap.yaml
nak3: Yes... I think it is a mistake and we should remove `strict_dns`.

nak3:
@skonto @maschmid Is it not OK to merge this?

maschmid:
Hi, I want to test it with this exact bits first, I didn't get to it last week.

PR #924, opened 2022-09-27
Remove deprecated empty fields in bootstrap.yaml

nak3:
### Changes

As mentioned in https://github.com/knative-sandbox/net-kourier/pull/920, there are some deprecated empty configurations are remained, which makes confusion.

This patch cleans up them.

- :broom: Update or clean up current behavior


skonto:
I guess there is no semantic difference between empty and non-existence.

PR #925, opened 2022-09-27
Upgrade envoyproxy/go-control-plane dependency from v0.10.1 to v0.10.3

tardieu:
# Changes

Upgrade `go-control-plane` dependency from v0.10.1 to v0.10.3 (latest).

This change is necessary to reflect and eventually make use of recent Envoy extensions such as `envoy.filters.http.stateful_session` that are available in Envoy v1.21 but not defined in `go-control-plane` v0.10.1.

This upgrade requires minor changes to three `kourier` source files due to the introduction of the `ResourceSnapshot` interface in https://github.com/envoyproxy/go-control-plane/commit/a51ba6fa54b2e107f5606a0e2799f8afed116c14 and the deprecation of `grpc.Insecure()`:
 - `cmd/kourier/health.go`
 - `pkg/envoy/server/xds_server.go`
 -  `pkg/generator/caches.go`

All other changes are contained in the `vendor` tree and obtained by running `./hack/update-deps.sh`.

/kind enhancement

**Release Note**

```release-note
Bump envoyproxy/go-control-plane version to v0.10.3
```


nak3:
/lgtm
/approve

Thank you so much!

PR #926, opened 2022-09-27
update envoy go-control-plane to v0.10.3

nader-ziada:
# Changes

- update envoy go-control-plane dependency to v0.10.3 in preparation for the update to the k8s api v0.25.2

**Release Note**

```release-note
update envoy go-control-plane dependency to v0.10.3
```


nader-ziada:
there is already another PR doing the update that I missed

/close

PR #934, opened 2022-10-13
Run envoy as non-root

skonto:
# Changes

- Runs envoy as non-root, similar to https://github.com/projectcontour/contour/pull/4084
- On OCP we use non-root for [some time](https://github.com/openshift-knative/serverless-operator/blob/1046d153bdd1bb4ba223386a3f54a2757b8d640c/openshift-knative-operator/cmd/operator/kodata/ingress/1.4/0-kourier.yaml#L430).
-  Tested on kind 1.25 with restricted PSP:
```
  kubectl create ns kourier-system
  kubectl label --overwrite ns kourier-system\
    pod-security.kubernetes.io/enforce=restricted \
    pod-security.kubernetes.io/warn=restricted \
    pod-security.kubernetes.io/audit=restricted
$ oc get pods -n kourier-system
NAME                                      READY   STATUS    RESTARTS   AGE
3scale-kourier-gateway-85878884dd-4wp6n   1/1     Running   0          16m
```

### DIFF
--- a/config/300-gateway.yaml
+++ b/config/300-gateway.yaml
@@ -68,7 +68,9 @@ spec:
           securityContext:
             allowPrivilegeEscalation: false
             readOnlyRootFilesystem: false
-            runAsNonRoot: false
+            runAsNonRoot: true
+            runAsUser: 65534
+            runAsGroup: 65534
             capabilities:
               drop:
                 - ALL


nak3:
Yeah, I think `runAsUser` is not allowed on OCP by default (or some condition).

The doc still mentions about it:
https://docs.openshift.com/container-platform/4.11/authentication/managing-security-context-constraints.html#security-context-constraints-example_configuring-internal-oauth
> A container or pod that requests a specific user ID will be accepted by OpenShift Container Platform only when a service account or a user is granted access to a SCC that allows such a user ID. The SCC can allow arbitrary IDs, an ID that falls into a range, or the exact user ID specific to the request.

With said, we (downstream) already tweak `runAsNonRoot` by a patch, so Stavros is going to patch the change (`runAsUser`) on downstream? (Or we don't care about it in the newer OCP version?)

nak3:
/lgtm
I assume that OCP will be solved somehow.

skonto:
@nak3 yeah probably we will patch it, it is not a problem. ;)

PR #936, opened 2022-10-14
Add private key provider support

### LINKED ISSUE #887: Feature Request Discussion: Enable configuring Envoy private key providers
# Introduction
As an ingress gateway, Kourier need handle a large number of incoming TLS and invoke Envoy to process them. Envoy supports configuring private key providers and the private key provider (CryptoMB) was released in 1.20 contrib build (https://github.com/envoyproxy/envoy/pull/17826 ).

CryptoMB private key provider allows Envoy to handle the incoming TLS handshakes’ RSA operations with the help of AVX-512 multi-buffer instructions, which allow up to eight RSA or ECDSA operations to be gathered into a buffer and processed simultaneously. This helps increase the performance especially when there are a large number of incoming TLS requests. 

Intel has published a whitepaper that contains more information about the AVX-512 instructions and potential performance increase: https://www.intel.com/content/www/us/en/architecture-and-technology/crypto-acceleration-in-xeon-scalable-processors-wp.html.
You can also find a presentation here: https://www.intel.com/content/www/us/en/developer/videos/accelerate-envoy-with-advanced-vector-extensions.html

To use the CryptoMB private key providers, users can simply configure Envoy like:
```
 tls_certificates:
   certificate_chain: { "filename": "/path/cert.pem" }
   private_key_provider:
     provider_name: cryptomb
     typed_config:
       "@type": type.googleapis.com/envoy.extensions.private_key_providers.cryptomb.v3alpha.CryptoMbPrivateKeyMethodConfig
       private_key: { "filename": "/path/key.pem" }       
       poll_delay: 0.1s
```
# Proposal
I think Kourier also needs to support this feature.  As an ingress gateway, it should be capable of handling such a large number of TLS at the same time. 
We should:
1. make the envoy configuration configurable, instead of using a static YAML file. 
2. add a configure option in Kourier's ingress gateway and let it pass the private key provider info to Envoy's configuration. 

Please kindly leave your comments if any suggestions or questions.

xxinran:
Fixes #887 

# Changes
* Add a new configuration option `enable-cryptomb` and set "false" by default.
* Update envoy configuration according to this configuration. If it is set to "true", Kourier need update envoy snapshot following the format [here](https://www.envoyproxy.io/docs/envoy/latest/api-v3/extensions/private_key_providers/cryptomb/v3alpha/cryptomb.proto#extensions-private-key-providers-cryptomb-v3alpha-cryptombprivatekeymethodconfig).

**Release Note**

This PR adds private key provider support in Kourier, which allow Kourier to improvement the TLS performance. 

For more information about cryptoMB private key provider, please refer to this [link](https://www.envoyproxy.io/docs/envoy/latest/api-v3/extensions/private_key_providers/cryptomb/v3alpha/cryptomb.proto#cryptomb-private-key-provider-proto).


xxinran:
@nak3  @lionelvillard

nak3:
/lgtm
/hold

`hold` for the CI. Let's merge this once CI passed. Thank you so much Xinran!

PR #958, opened 2022-12-07
Bump Maistra image version to 2.3

nak3:
This patch bumps maistra (Red Hat's Envoy image) to 2.3 in github action.

### DIFF
--- a/.github/workflows/kind-e2e.yaml
+++ b/.github/workflows/kind-e2e.yaml
@@ -24,7 +24,7 @@ jobs:
         - v1.25.3
 
         gateway:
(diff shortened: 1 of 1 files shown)


dprotaso:
@nak3 looks like the tls tests are blocking

nak3:
/retest
Thank you. Yeah, I will look into it.

nak3:
I found the root cause but need some time to change the code.
Just in case, the upgrade test does not use the maistra image so this PR is not related to the failure.

dprotaso:
/override KinD e2e upgrade tests / e2e tests (v1.24.7, tls) 

/lgtm
/approve

PR #959, opened 2022-12-08
Determine internal-encryption via KIngress 

nak3:
This patch changes controller to determine the internal-encryption via
KIngress's port number and service's port name.

The configuration is prefilled by the "default" values so it creates an
unexpected envoy config when users set non-default value in
`config-network` such as `internal-encryption`.


pkg/generator/ingress_translator_test.go
```diff
@@ -907,153 +907,18 @@ func TestIngressTranslatorWithUpstreamTLS(t *testing.T) {
...
-						v3.Cluster_STATIC,
-					),
-				},
-				externalVirtualHosts:    vHosts,
-				externalTLSVirtualHosts: []*route.VirtualHost{},
-				internalVirtualHosts:    vHosts,
-			}
-		}(),
-	}, {
-		name: "http and https",
```
nak3, line 971: These tests are same with `simple` and `http2` test.
Serving always has two ports `http` and `https` so the `svc()` func returns service with the two ports.

nak3:
@skonto @dprotaso could you please take a look?

pkg/generator/ingress_translator.go
```diff
@@ -192,13 +196,17 @@ func (translator *IngressTranslator) translateIngress(ctx context.Context, ingre
 
 				var transportSocket *envoycorev3.TransportSocket
 
-				// This has to be "OrDefaults" because this path is called before the informers are
+				// This has to be "OrDefaults" because this path could be called before the informers are
```
skonto, line 199: Could we just avoid setting up the envoy configuration too early eg. when creating a new controller? For example right now we expect each controller to set is own cm watches, call Watch on the passed cmw for each cmw and then later on we Start() the cmw.
I propose to solve this once and for all and do either:
a) allow a controller to have a postCreate func to be called after cmw is started (if that is useful as in Kourier).
b) Allow starting cmw in the constructor and skip it at shared main.
The only thing you need to be careful is not to start it too early as you will hit issues like [this](https://github.com/knative/eventing/pull/6615/files#diff-2a10da0d0184623583740c11a53b98f7cb293507b027dfa19622e02e08ef4e10).

nak3: Avoiding to set up the envoy configuration is also fine for me. But I am not so familiar with controller stuff so I would like to ask you to send the PR.

skonto: The easiest way would be to fetch the cm imho. Let's see what @dprotaso has to say. If we want to add flexibility at sharedmain I can send a PR for sure.

skonto: @nak3 check here: https://github.com/knative/serving/blob/3812acfe03bce45acd332c7706ff1564446425cb/cmd/autoscaler/main.go#L129

skonto:
LGTM
@nak3 could you add a TODO comment to remove this if we add cm prefetching in sharedmain as activator/autoscalers do. I can then stamp it.

nak3:
Okay, added TODO comment.

nak3:
@skonto Could you please approve this?

nak3:
@skonto I think all test passed.

PR #960, opened 2022-12-08
Rebuild using Go 1.19.4 (to address CVEs)

dprotaso:
/assign @psschwei 

cc @nak3


dprotaso:
wrong target branch will make a new one

PR #969, opened 2022-12-16
Add upgrade strategy

nak3:
As per title, this patch adds upgrade strategy.

Downstream has been using [this strategy](https://github.com/openshift-knative/net-kourier/blob/release-1.8/openshift/release/download_release_artifacts.sh#L36-L38) and verified that it becomes stable.


nak3:
@skonto Could you approve this?

evankanderson (commented):
/approve

PR #976, opened 2023-01-20
Bump envoy image version to 1.22 and 1.23

nak3:
As per tittle, this patch bumps envoy testing version to 1.22 and 1.23.


evankanderson (commented):
/lgtm
/approve

PR #978, opened 2023-01-21
Add prometheus annotations to net-kourier

### LINKED ISSUE #957: Add Prometheus monitoring annotations to `envoy`, for the cases where Prometheus is used.
From a 1:1 Slack conversation on the CNCF slack with @aryan9600:

> I wanted to ask if its possible to make a small change to the default knative installation.
> since, the whole point of flagger is to do progressive delivery in a metrics driven way, it’d be great if the envoy deployment that is deployed in the kourier-system ns, could have the prometheus annotations as well? that way all the envoy metrics can be automatically scraped by prometheus. if there’s a way to specify this at install time, please let me know. its a bit hacky to tell users to add annotations to the pod template of a deployment, so i wanna keep that as a last resort.

I believe that Sanskar is asking that we add the following annotations to the Pod template (which I can't easily Google, apparently!):

```yaml
spec:
  template:
    metadata:
      annotations:
        prometheus.io/scrape: "true"
        prometheus.io/port: "8002"
        prometheus.io/path: "/stats/prometheus"
```

/assign @evankanderson

evankanderson:
# Changes

Fixes #957 
- 🎁  Adds prometheus scrape annotations for both `controller` and `envoy` pods.

**Release Note**

```release-note
* Adds prometheus scraping annotations for net-kourier, so Prometheus Operator users should now be able to collect statistics automatically.
```

### DIFF
--- a/config/300-controller.yaml
+++ b/config/300-controller.yaml
@@ -34,6 +34,10 @@ spec:
       app: net-kourier-controller
   template:
     metadata:
+      annotations:
+        prometheus.io/scrape: "true"
+        prometheus.io/port: "9090"
+        prometheus.io/path: "/metrics"
       labels:
         app: net-kourier-controller
     spec:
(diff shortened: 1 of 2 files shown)


evankanderson:
/assign @nak3 @psschwei @skonto 

cc @aryan9600

(you can `/lgtm` even if you're not in the org if you want)

Issue #923, opened 2022-09-26
Use k8s's default gRPC readinessProbe 


nak3:
Currently, readinessProbe uses the custom command like https://github.com/knative-sandbox/net-kourier/blob/main/config/300-controller.yaml#L57-L59
It is just same command with https://github.com/grpc-ecosystem/grpc-health-probe so I think it is possible to replace with the k8s's default gRPC command.

Since 1.24, the feature is beta - https://kubernetes.io/blog/2022/05/13/grpc-probes-now-in-beta/ and it should be okay to start using it.

PR #981, opened 2023-01-25
Use default gRPC readiness probe

nak3:
As per title, this patch changes to use default gRPC readiness probe.

__Additional Note__
Current timeout is 100 milli seconds but `timeoutSeconds` is minimum 1 second so this change will make the timeout longer than current.

```
$ kubectl explain pod.spec.containers.readinessProbe.timeoutSeconds
KIND:     Pod
VERSION:  v1

FIELD:    timeoutSeconds <integer>

DESCRIPTION:
     Number of seconds after which the probe times out. Defaults to 1 second.
     Minimum value is 1. More info:
     https://kubernetes.io/docs/concepts/workloads/pods/pod-lifecycle#container-probes
```

Fix https://github.com/knative-sandbox/net-kourier/issues/923


nak3:
This will be a breaking change for the old k8s cluster (or OCP) so we will not merge until we are sure that the gRPC readiness probe is available. (after 1.10 or 1.11 release.)
But please feel free to ping here if somebody wants to get this PR merge. Downstream can make a patch if necessary.

nak3:
/hold cancel
/cc @skonto @ReToCode 

This is now ready.

ReToCode (approved):
Nice, I like it.

/lgtm
/hold for other reviewers to chime in.

PR #1025, opened 2023-04-11
Fix keepalive interval

ryutoyasugi:
# Changes
- :bug: Extend keepalive interval.

/kind bug

# Why
After updating kourier from v1.5.0 to v1.9.2, warning logs appeared and `envoy_cluster_manager_cds_update_failure` was counted.
```
kourier-gateway [2023-04-04 05:06:31.483][1][warning][config] [./source/common/config/grpc_stream.h:160] StreamAggregatedResources gRPC config stream closed: 13,
```
I extended the keepalive interval to 5 min and it resolved.
refs: #920 

This setting may be related.
https://pkg.go.dev/google.golang.org/grpc/keepalive#EnforcementPolicy
```go
MinTime time.Duration // The current default value is 5 minutes.
```

### DIFF
--- a/config/200-bootstrap.yaml
+++ b/config/200-bootstrap.yaml
@@ -93,7 +93,7 @@ data:
               explicit_http_config:
                 http2_protocol_options:
                   connection_keepalive:
-                    interval: 30s
+                    interval: 300s
                     timeout: 5s
           connect_timeout: 1s
           load_assignment:
--- a/config/200-bootstrap.yaml
+++ b/config/200-bootstrap.yaml
@@ -93,7 +93,7 @@ data:
               explicit_http_config:
                 http2_protocol_options:
                   connection_keepalive:
-                    interval: 300s
+                    interval: 30s
                     timeout: 5s
           connect_timeout: 1s
           load_assignment:
--- a/pkg/envoy/server/xds_server.go
+++ b/pkg/envoy/server/xds_server.go
@@ -29,6 +29,7 @@ import (
 	xds "github.com/envoyproxy/go-control-plane/pkg/server/v3"
 	"google.golang.org/grpc"
 	health "google.golang.org/grpc/health/grpc_health_v1"
+	"google.golang.org/grpc/keepalive"
 )
 
 const (
@@ -69,7 +70,10 @@ func (envoyXdsServer *XdsServer) RunManagementServer() error {
 	port := envoyXdsServer.managementPort
 	server := envoyXdsServer.server
 
-	grpcServer := grpc.NewServer(grpc.MaxConcurrentStreams(grpcMaxConcurrentStreams))
+	grpcServer := grpc.NewServer(
+		grpc.MaxConcurrentStreams(grpcMaxConcurrentStreams),
+		grpc.KeepaliveEnforcementPolicy(keepalive.EnforcementPolicy{PermitWithoutStream: true}),
+	)
 	lis, err := net.Listen("tcp", fmt.Sprintf(":%d", port))
 	if err != nil {
 		return fmt.Errorf("failed to listen: %w", err)
--- a/pkg/envoy/server/xds_server.go
+++ b/pkg/envoy/server/xds_server.go
@@ -20,6 +20,7 @@ import (
 	"context"
 	"fmt"
 	"net"
+	"time"
 
 	cluster "github.com/envoyproxy/go-control-plane/envoy/service/cluster/v3"
 	discovery "github.com/envoyproxy/go-control-plane/envoy/service/discovery/v3"
@@ -72,7 +73,7 @@ func (envoyXdsServer *XdsServer) RunManagementServer() error {
 
 	grpcServer := grpc.NewServer(
 		grpc.MaxConcurrentStreams(grpcMaxConcurrentStreams),
-		grpc.KeepaliveEnforcementPolicy(keepalive.EnforcementPolicy{PermitWithoutStream: true}),
+		grpc.KeepaliveEnforcementPolicy(keepalive.EnforcementPolicy{MinTime: 20 * time.Second, PermitWithoutStream: true}),
 	)
 	lis, err := net.Listen("tcp", fmt.Sprintf(":%d", port))
 	if err != nil {


nak3:
Thank you so much for this PR. I agree that the root cause you found out.

While it makes sense to increase the interval, we (downstream) experienced an issue without https://github.com/knative-sandbox/net-kourier/pull/920, which Kourier Gateway does not get the updated config, so I am worried that the 300s is too long. In addition, we always have to set the `interval` more than 300s.

So, how do you think about setting `PermitWithoutStream` to `true` on the server side - i.e. `pkg/envoy/server/xds_server.go`.

ryutoyasugi:
I too think it's good to set `PermitWithoutStream: true`.
https://github.com/knative-sandbox/net-kourier/blob/release-1.9/pkg/envoy/server/xds_server.go#L72
```go
	grpcServer := grpc.NewServer(
		grpc.MaxConcurrentStreams(grpcMaxConcurrentStreams),
		grpc.KeepaliveEnforcementPolicy(keepalive.EnforcementPolicy{PermitWithoutStream: true}),
	)
```
If there are no problems, I will fix it in this pull request.

nak3:
Sounds great! Let's add the fix you suggested https://github.com/knative-sandbox/net-kourier/pull/1025#issuecomment-1504552383. Could you update here or send the PR?

skonto:
```
// If true, client sends keepalive pings even with no active RPCs. If false,
// when there are no active RPCs, Time and Timeout will be ignored and no
// keepalive pings will be sent.
PermitWithoutStream bool // false by default.
```

This was meant to protect the server from malicious clients. 

Regarding the policy:

```
// EnforcementPolicy is used to set keepalive enforcement policy on the
// server-side. Server will close connection with a client that violates this
// policy.
type EnforcementPolicy struct {
	// MinTime is the minimum amount of time a client should wait before sending
	// a keepalive ping.
	MinTime time.Duration // The current default value is 5 minutes.
	// If true, server allows keepalive pings even when there are no active
	// streams(RPCs). If false, and client sends ping when there are no active
	// streams, server will send GOAWAY and close the connection.
	PermitWithoutStream bool // false by default.
```
}

Here we increase the keep alive at the gateway side to 5min and allow pings from gateway to XDs (net-kourier) even if the connection is idle.
a) Is not the case that the client (gateway) needs to have the same setting for this to work check [here](https://github.com/grpc/grpc-go/issues/2325#issuecomment-425559715)? Does this happen? What is the behavior?
I suspect that since minTime is 5min and we send every 5min we probably dont notice and ping denial.
b) Increasing the keepalive means we send a ping every 5min from the gateway to confirm that the server is up. I am wondering what happened within the 5min period and why the ping with the 30s period failed?
I am wondering if you experienced some transient errors and a higher value for the timeout at the keepalive settings (default is  timeout: 5s) would be better.

In any case I would like to understand better the behavior before merging.

config/200-bootstrap.yaml
```diff
@@ -93,7 +93,7 @@ data:
               explicit_http_config:
                 http2_protocol_options:
                   connection_keepalive:
-                    interval: 30s
+                    interval: 300s
```
nak3, line 96: Hmm... I thought that we don't need to increase the interval as we set `PermitWithoutStream` to true. Do we still need this change?

ryutoyasugi:
@skonto Thank you for the question.
> a) Is not the case that the client (gateway) needs to have the same setting for this to work check https://github.com/grpc/grpc-go/issues/2325#issuecomment-425559715? Does this happen? What is the behavior?

If the client has not set `PermitWithoutStream: true`, I think keepalive fail when no active stream.

> b) Increasing the keepalive means we send a ping every 5min from the gateway to confirm that the server is up. I am wondering what happened within the 5min period and why the ping with the 30s period failed?

Pings sent within `EnforcementPolicy.MinTime` always fail. Extending the timeout didn't change.

config/200-bootstrap.yaml
ryutoyasugi: `PermitWithoutStream` is a setting that allows ping when there is no active stream.
So it is necessary to extend interval to `EnforcementPolicy.MinTime`.

ryutoyasugi:
@nak3 @skonto 
I think that setting `EnforcementPolicy.MinTime` to 30s will also solve it.
Then keepalive interval can be left at 30s.

https://github.com/knative-sandbox/net-kourier/blob/release-1.9/pkg/envoy/server/xds_server.go#L72
```go
	grpcServer := grpc.NewServer(
		grpc.MaxConcurrentStreams(grpcMaxConcurrentStreams),
		grpc.KeepaliveEnforcementPolicy(keepalive.EnforcementPolicy{MinTime: 30 * time.Second}),
	)
```

nak3:
Setting `EnforcementPolicy.MinTime` to `30s` sounds good to me. @skonto Do you have any objection?

skonto:
I double check this is indeed an issue I get with current settings:

```
[2023-04-17 10:01:30.546][1][warning][config] [./source/common/config/grpc_stream.h:160] StreamAggregatedResources gRPC config stream closed: 13, 
```

I looked at the grpc go lib so here are the parts we care about when you set no policy by default:

https://github.com/grpc/grpc-go/blob/master/internal/transport/http2_server.go#L246-L247
```
	kep := config.KeepalivePolicy
	if kep.MinTime == 0 {
		kep.MinTime = defaultKeepalivePolicyMinTime
	}
```

If we dont set anything it gets 5min so 30s < 5min.

https://github.com/grpc/grpc-go/blob/master/internal/transport/http2_server.go#L903-L914

```
	if ns < 1 && !t.kep.PermitWithoutStream {
		// Keepalive shouldn't be active thus, this new ping should
		// have come after at least defaultPingTimeout.
		if t.lastPingAt.Add(defaultPingTimeout).After(now) {
			t.pingStrikes++
		}
	} else {
		// Check if keepalive policy is respected.
		if t.lastPingAt.Add(t.kep.MinTime).After(now) {
			t.pingStrikes++
		}
	}
```

So I think we should indeed set PermitWithoutStream=1 to avoid comparing with defaultPingTimeout which is set to 2hours. then it means we need to set minTime to <= 30sec.

Ps. I am off this week but I don't want to block you.

ryutoyasugi:
@skonto 
Thank you for your help!!
I fixed this PR.

nak3:
/lgtm
/approve

Thank you!

PR #1029, opened 2023-04-13
Add missing probes

skonto:
# Changes

- Adds missing probes for net-kourier-controller and the gateway,
- changes the path to "/" so it can be used by both paths. Adding a separate path per probe can be more complex due to the cluster matching etc.

### DIFF
--- a/pkg/generator/status_vhost.go
+++ b/pkg/generator/status_vhost.go
@@ -28,7 +28,7 @@ import (
 )
 
 // Generates an internal virtual host that signals that the Envoy instance has
-// been configured, this endpoint is used by the kubernetes readiness probe.
+// been configured, this endpoint is used by the kubernetes readiness, liveness probes.
 func statusVHost() *route.VirtualHost {
 	vhost := envoy.NewVirtualHost(
 		config.InternalKourierDomain,
@@ -54,7 +54,7 @@ func readyRoute() *route.Route {
 	cluster := envoy.NewWeightedCluster("service_stats", 100, nil)
 	var wrs []*route.WeightedCluster_ClusterWeight
 	wrs = append(wrs, cluster)
-	route := envoy.NewRoute("gateway_ready", nil, "/ready", wrs, 1*time.Second, nil, "")
+	route := envoy.NewRoute("gateway_ready", nil, "/", wrs, 1*time.Second, nil, "")
 
 	return route
 }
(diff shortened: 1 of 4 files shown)


ReToCode (approved):
/lgtm
/hold for @nak3 to review.

pkg/generator/status_vhost.go
```diff
@@ -54,7 +54,7 @@ func readyRoute() *route.Route {
 	cluster := envoy.NewWeightedCluster("service_stats", 100, nil)
 	var wrs []*route.WeightedCluster_ClusterWeight
 	wrs = append(wrs, cluster)
-	route := envoy.NewRoute("gateway_ready", nil, "/ready", wrs, 1*time.Second, nil, "")
+	route := envoy.NewRoute("gateway_ready", nil, "/", wrs, 1*time.Second, nil, "")
```
nak3, line 57: Should we use `/probe`, `/kprobe`, `/health` or `/khealth` or something else for both readiness and liveness?
Using the root path should be fine as long as the header matching is used but I am not 100% confident if it is no side effect at all. (e.g. the total number of access to `/` cannot be counted anymore.)

skonto:
@nak3 I update the path. Sp both readiness and liveness probes use the /ready path at the admin interface.
I dont think there is anything else to use there (it is the same path to check envoy liveness that you get at the admin ui).

I tried to use a different path like "/health" by doing: 
- add the subpath to the regex in the boostrap.yaml
- create a new route with:
```
eroute.Action = &route.Route_Redirect{
  Redirect: &route.RedirectAction{
    PathRewriteSpecifier: &route.RedirectAction_PathRedirect{
      PathRedirect: "/ready",
    },
  },
}
```
and although K8s allows redirects: https://github.com/kubernetes/kubernetes/blob/master/pkg/probe/http/http.go#L113
I get a warning from that line:
$ k get events -n kourier-system | grep warning
4m46s       Warning   ProbeWarning                   pod/3scale-kourier-gateway-6c9ddb7b68-kgjhd      Liveness probe warning: Probe terminated redirects, Response body:
So although it work due to the warning I decided not to go that path.

nak3:
Thank you. My intention was not used the same path (`/`) with applications so `/ready` should be fine. Let's wait for serving's PR if there is any additional feedback if we can apply.

dprotaso:
FYI Serving's PR merged

nak3:
/lgtm
/hold cancel

Thank you!

PR #831, opened 2022-04-26
Fixes #830

sidharthramesh:
I've just added a very minor change that will fix #830.
Haven't written any tests for this. I'm not great with writing tests in Go and might need more help here.
I've set the default value of `KOURIER_EXTAUTHZ_PACKASBYTES` to false so that it doesn't break any existing applications, so I'm expecting all the existing tests to pass 🤞.

In the future, I think that having a mechanism to pass such configuration parameters directly to Envoy would be great.

PR #1039, opened 2023-05-06
Add support for pack_as_bytes configuration for ext-authz

### LINKED ISSUE #830: Add support for pack_as_bytes configuration for ext_authz.v3
Hey there!
I've been forwarding an internal ArgoCD through a Knative Service (caddy proxy) to expose it to the internet.
I'm using Kourier along with External Authorization (using OPA)

Today I was trying to execute an ArgoCD CLI command: `argocd cluster add <context>` and it failed with a 403 error, with no logs on OPA. There was no rule defined to disallow this request btw.

After hours of searching through logs, I found that when the Argocd CLI tries to make a grpc web request, and this happens in Kourier:
```
[libprotobuf ERROR external/com_google_protobuf/src/google/protobuf/wire_format_lite.cc:577] String field 'envoy.service.auth.v3.AttributeContext.HttpRequest.body' contains invalid UTF-8 data when serializing a protocol buffer. Use the 'bytes' type if you intend to send raw bytes. 

[2022-04-26T16:34:13.257Z] "POST /cluster.ClusterService/Create HTTP/1.1" 403 UAEX 2239 0 1 - "-" "argocd-client/v2.3.3+07ac038.dirty grpc-go/1.15.0" "cf34b22f-2a40-427a-bdd6-a7c9987e92e3" "argocd.dv.kn.voiceintern.com:443" "-"
```
Basically Envoy is trying to convert every request body into a UTF-8 string. This will not be possible in all cases and will fail on [many edge cases](https://github.com/envoyproxy/envoy/issues/9822).

The good news is that the [`pack_as_bytes` parameter]() enables Envoy to send the request as raw bytes without converting into a UTF-8 string: https://github.com/envoyproxy/envoy/pull/13131

This has also been documented in more detail here: https://www.envoyproxy.io/docs/envoy/latest/api-v3/extensions/filters/http/ext_authz/v3/ext_authz.proto.html?highlight=pack_as_bytes#extensions-filters-http-ext-authz-v3-buffersettings

Proposal: Add a `KOURIER_EXTAUTHZ_PACKASBYTES` environment variable to allow passing the `pack_as_bytes` parameter to Envoy.

norbjd:
Supersedes #831. This #831 PR is opened for more than 1 year as of today, and making changes on it introduces a lot of delay, since I'm not the author and I need to synchronize with the original PR creator (I've made 3 PR there yet, and when CI tests are failing, I need to create another MR every time :disappointed:). I am currently running a fork of Kourier in production just for this change, so I'd like this to be merged upstream soon, and it can also profit other users. Thanks!

Credits also go to [sidharthramesh](https://github.com/sidharthramesh) who reported the issue and made the initial PR. I've added the tests (unit + integration), fix lint issues, doc, etc.

# Changes

:gift: Add support for pack_as_bytes configuration for ext-authz

When ext-authz uses grpc protocol, it's now possible to configure pack_as_bytes. This is necessary when the body sent to the external authorization service is raw bytes, and not an UTF-8 string.

By default, this value is set to `false`, so there is no breaking changes.

If users try to set `KOURIER_EXTAUTHZ_PACKASBYTES` to `true` if they are using the HTTP protocol for ext-authz, kourier will throw an error at startup. Indeed, pack as bytes option makes only sense when gRPC protocol is used.

/kind enhancement

Fixes #830

**Release Note**

```release-note
add support for pack_as_bytes configuration for ext-authz
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -151,7 +151,11 @@ vars in the `net-kourier-controller` deployment:
 - `KOURIER_EXTAUTHZ_PATHPREFIX`: If `KOURIER_EXTAUTHZ_PROTOCOL` is equal to
   http or https, path to query the ext auth service. Example : if set to
   `/verify`, it will query `/verify/` (**notice the trailing `/`**).
-  If not set, it will query `/`.
+  If not set, it will query `/`
+- `KOURIER_EXTAUTHZ_PACKASBYTES`: If `KOURIER_EXTAUTHZ_PROTOCOL` is equal to
+  grpc, sends the body as raw bytes instead of a UTF-8 string. Accepts only true/false, t/f or 1/0.
+  Attempting to set another value will throw an error. Defaults to false. More info
+  [Envoy Docs](https://www.envoyproxy.io/docs/envoy/latest/api-v3/extensions/filters/http/ext_authz/v3/ext_authz.proto.html?highlight=pack_as_bytes#extensions-filters-http-ext-authz-v3-buffersettings).
 
 `*` Required
 
--- a/pkg/config/ext_authz.go
+++ b/pkg/config/ext_authz.go
@@ -17,6 +17,7 @@ limitations under the License.
 package config
 
 import (
+	"errors"
 	"fmt"
 	"net"
 	"os"
@@ -78,6 +79,7 @@ type config struct {
 	MaxRequestBytes  uint32           `default:"8192"`
 	Timeout          int              `default:"2000"`
 	Protocol         extAuthzProtocol `default:"grpc"`
+	PackAsBytes      bool             `default:"false"`
 	PathPrefix       string
 }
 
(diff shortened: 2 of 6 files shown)


nak3:
@norbjd Thank you so much for this and your investigation! It is very clear :pray: 
The failure is 100% Envoy bug and we know it will be fixed in the newer version so I think we should skip the test for maistra.
Can we introduce an env value to distinguish maistra image and insert the `skip` code into the e2e this?

```go
        if maistra = os.Getenv(some_new_env_to_distinguish_maistra); maistra != "" {
          t.Skip("Skip due to issue #xxx") // TODO: Drop this skip after maistra started using Envoy 1.22.10 or later.
        }
```

nak3:
/lgtm
/approve

Thank you so much!

PR #1040, opened 2023-05-08
Fix Secret and Service tracker

nak3:
Currently controller has a tracker for Secret and Service but it is not triggered due to wrong plural (`s`).

This patch fixes it and make the tracker working when Secret or Service are changed.

```release-note
There is a bug which tracker was not triggered for Secret and Service. It is fixed now.
```

### DIFF
--- a/pkg/reconciler/ingress/controller.go
+++ b/pkg/reconciler/ingress/controller.go
@@ -266,7 +266,7 @@ func NewController(ctx context.Context, cmw configmap.Watcher) *controller.Impl 
 	serviceInformer.Informer().AddEventHandler(controller.HandleAll(
 		controller.EnsureTypeMeta(
 			impl.Tracker.OnChanged,
-			corev1.SchemeGroupVersion.WithKind("Services"),
+			corev1.SchemeGroupVersion.WithKind("Service"),
 		),
 	))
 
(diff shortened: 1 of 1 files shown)


pkg/reconciler/ingress/controller.go
```diff
@@ -266,7 +266,7 @@ func NewController(ctx context.Context, cmw configmap.Watcher) *controller.Impl
 	serviceInformer.Informer().AddEventHandler(controller.HandleAll(
 		controller.EnsureTypeMeta(
 			impl.Tracker.OnChanged,
-			corev1.SchemeGroupVersion.WithKind("Services"),
+			corev1.SchemeGroupVersion.WithKind("Service"),
```
skonto, line 269: Interesting. What about Endpoints and other resources?

skonto:
Nice catch!

pkg/reconciler/ingress/controller.go
nak3: Endpoints is no problem.

pkg/reconciler/ingress/controller.go
```diff
@@ -293,7 +293,7 @@ func NewController(ctx context.Context, cmw configmap.Watcher) *controller.Impl
 	secretInformer.Informer().AddEventHandler(controller.HandleAll(
 		controller.EnsureTypeMeta(
 			impl.Tracker.OnChanged,
-			corev1.SchemeGroupVersion.WithKind("Secrets"),
+			corev1.SchemeGroupVersion.WithKind("Secret"),
```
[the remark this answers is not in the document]
nak3: The `Kind` is literally Kubernete resource's kind e.g.: 
```
$ kubectl  get secrets  -o yaml |grep kind:
  kind: Secret
```
So it makes sense to be same with `ingress_translator.go`, but actually the name is Kubernete global and I think we don't need to introduce the new value for the consistency in net-kourier's code.

In addition, I grep-ed in serving code with `$ grep -irH WithKind ${SERVING_CODE}` but all of the codes use the `"Secret"` or such kind of the fixed value.

Issue #1015, opened 2023-03-27
Test with envoy:v1.25 


nak3:
It seems envoy 1.25 has a change which brakes our test so we need to fix it.

log - https://github.com/knative-sandbox/net-kourier/actions/runs/4528728473/jobs/7975777700?pr=1014

PR #1047, opened 2023-05-11
Drop envoy 1.23 and start testing 1.25

nak3:
Fix https://github.com/knative-sandbox/net-kourier/issues/1015

Envoy 1.23 will be EOL in a few months - https://github.com/envoyproxy/envoy/blob/main/RELEASES.md#major-release-schedule
This patch drops 1.23 testing and use 1.24 as the default image. Then, start testing 1.25.

/cc @skonto @ReToCode @norbjd

### DIFF
--- a/test/timeout/timeout_test.go
+++ b/test/timeout/timeout_test.go
@@ -62,20 +62,18 @@ func TestIdleTimeout(t *testing.T) {
 
 	test := struct {
 		name         string
-		code         int
 		initialDelay time.Duration
 		delay        time.Duration
 	}{
 		name:         "100s delay before response",
-		code:         http.StatusRequestTimeout,
 		initialDelay: waitDuration,
 	}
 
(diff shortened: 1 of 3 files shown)


test/timeout/timeout_test.go
```diff
@@ -87,5 +85,5 @@ func checkTimeout(t *testing.T, client *http.Client, name string, code int, init
 		t.Fatal(err)
 	}
 	defer resp.Body.Close()
-	assert.Equal(t, resp.StatusCode, code)
+	assert.True(t, resp.StatusCode == http.StatusRequestTimeout || resp.StatusCode == http.StatusGatewayTimeout)
```
nak3, line 88: The status code was changed by Envoy upstream - https://github.com/knative-sandbox/net-kourier/issues/1015#issuecomment-1542258326

I considered to use `KOURIER_GATEWAY_IMAGE` to distinguish the version but it is a minor stuff so I think it is alright for test with "either" match.

PR #1058, opened 2023-06-06
Set TLS 1.3 for min and max versions

nak3:
This patch make a change that traffic from Kourier to Activator/QP uses TLS 1.3
when internal encryption is enabled.

Due to an issue of Envoy https://github.com/envoyproxy/envoy/issues/9300,
we need to set TLS 1.3 for min and max version.

Part of https://github.com/knative/serving/issues/14057

**Release Note**

```release-note
Traffic from Kourier to Activator/QP uses TLS 1.3 when internal encryption is enabled.
```

/cc @dprotaso @skonto @ReToCode


nak3:
Verified on serving side:

* https://github.com/nak3/serving/actions/runs/5183705185 (current serving code)
* https://github.com/nak3/serving/actions/runs/5183817244 (set minTLS to 1.3 on serving side https://github.com/nak3/serving/pull/59/commits/7b175abfa55b4e4c4e26e30eaac40773e4fe4d4a)

ReToCode (approved):
/lgtm

PR #1063, opened 2023-06-14
Add startup probe for controller io mitigate probe failures caused by the long startup duration

nak3:
When there are many Kingress is deployed on the system, it takes long time to start net-kourier-controller (e.g. 12 minutes for 1000 ksvc) and so Liveness/Readiness probe fails.

We should improve the process but for now we add a startup probe (with long time 20 min for enough safety to start) to avoid the Liveness/Readiness probe failure.

**Release Note**

```release-note
NONE
```

EDIT: Delete release-note as it is reverted.

### DIFF
--- a/config/300-controller.yaml
+++ b/config/300-controller.yaml
@@ -63,6 +63,11 @@ spec:
           - name: http2-xds
             containerPort: 18000
             protocol: TCP
+          startupProbe:
+            grpc:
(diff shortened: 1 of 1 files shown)


ReToCode (approved):
/lgtm

Nice. We should (additionally) also look into making it faster in the first place next :)

Issue #968, opened 2022-12-09
Do not use `FromContextOrDefaults()` in `translateIngress()`


nak3:
Currently, there are some places use `FromContextOrDefaults()` in `translateIngress()` function such as:

https://github.com/knative-sandbox/net-kourier/blob/649db2781d60dd6a5580806194b3df5378118ed9/pkg/generator/ingress_translator.go#L195
https://github.com/knative-sandbox/net-kourier/blob/649db2781d60dd6a5580806194b3df5378118ed9/pkg/generator/ingress_translator.go#L268

`FromContextOrDefaults()` returns the default value in `config-network` so controller temporarily creates a envoy configuration with unexpected configuration.
For example, `internal-encryption` is disabled by default so it creates envoy configuration without `internal-encryption` even if users enabled.

ReToCode:
@nak3 do you know why we generate an initial envoy snapshot when the controller is constructed, instead just doing that on the first reconcile event? Is that so envoy can start?

nak3:
Sorry, I am not familiar with it but I think it was implemented as there is an issue that controller sends an incomplete configuration to an already running gateway container as per the following PRs.
- https://github.com/knative-sandbox/net-kourier/pull/72
- https://github.com/knative-sandbox/net-kourier/pull/137
- https://github.com/knative-sandbox/net-kourier/pull/276

PR #1066, opened 2023-06-21
Increase default QPS and Burst value to improve slow startup duration.

nak3:
:bug: This patch increases default QPS and Burst value to improve slow startup duration.

**Release Note**

```release-note
The default `kube-api-qps` and `kube-api-burst` are set to `200`. You can adjust them via env variable if necessary.
```

/cc @skonto @maschmid

### DIFF
--- a/config/300-controller.yaml
+++ b/config/300-controller.yaml
@@ -49,6 +49,10 @@ spec:
               value: ""
             - name: CERTS_SECRET_NAME
               value: ""
+            - name: KUBE_API_BURST
+              value: "200"
+            - name: KUBE_API_QPS
+              value: "200"
             - name: SYSTEM_NAMESPACE
               valueFrom:
                 fieldRef:


dprotaso:
Do you have a linked issue describing the slow startup?

maschmid:
There was originally  https://github.com/knative-sandbox/net-kourier/pull/879 , which was eventually closed as stale.

dprotaso:
Out of curiosity which outbound calls to the K8s API server are being throttled?

dprotaso:
> I believe it was services and endpoints gets, done by the IngressTranslator

Is there a reason those calls aren't using informers?

nak3:
I was not involved this discussion in https://github.com/knative-sandbox/net-kourier/pull/276 but the client correctly list all resources at startup.

https://github.com/knative-sandbox/net-kourier/blob/790358f73d01d37c908df5c7c54a7e275c22219e/pkg/reconciler/ingress/controller.go#L213-L214

I researched the context before as https://github.com/knative-sandbox/net-kourier/issues/968#issuecomment-1427260209 but it was a little bit complicated and still open https://github.com/knative-sandbox/net-kourier/issues/968 (it didn't intend to solve this informer vs client but I think related) for that.

skonto:
Finding good default values is an open question to me (depends on the scale we support). If we can't avoid the direct K8s api server calls then I don't see many options here. In general this is a problem faced elsewhere too see: https://github.com/kedacore/keda/issues/3730 & the fix https://github.com/kedacore/keda/pull/3731.

nak3:
I think it's not necessary to exert effort in finding the good default values as users can easily adjust them.
And we will actually turn off them in the future https://github.com/knative/pkg/issues/2756 so bigger value like `200` should be alright for now, I think.

nak3:
@skonto @dprotaso 

Could you merge this before 1.11 cut if you are alright?

nak3:
Yes, that's correct. The one "the listing to fill the informer cache is the slow piece right now", I need to test and confirm it by myself but I believe there is no change since https://github.com/knative-sandbox/net-kourier/pull/276.

skonto:
Afaik `getReadyIngresses` is the slow part that lists ingresses (with a big number things get stuck). This call access the api server directly https://github.com/knative-sandbox/net-kourier/blob/main/pkg/reconciler/ingress/controller.go#L317.

@nak3 could you confirm the last bits so we can merge.

nak3:
`getReadyIngresses` also calls k8s client but the critical part is `startupTranslator` which lists services, endpoints and etc as @maschmid answered above https://github.com/knative-sandbox/net-kourier/pull/1066#issuecomment-1604316418

Please see the log in https://github.com/knative-sandbox/net-kourier/issues/840 or (SRVKS-1078 our internal issue). It prints `Priming the config with ...` so `getReadyIngresses` passed without stuck.

skonto:
Yeah correct now I remember ;) Ok so should we merge? Are there any objections?

dprotaso:
I'm confused why doesn't `startupTranslator` use an informer?

nak3:
Hmm... I tested https://github.com/knative-sandbox/net-kourier/pull/1075. It works and the start speed is fast on 200 ksvcs.

I'm not sure what issue observed with the informer when https://github.com/knative-sandbox/net-kourier/pull/276 was created (the comment line clearly mentions `The startup translator uses clients instead of listeners to correctly list all resources at startup.` but it works on the current net-kourier.

Should we merge https://github.com/knative-sandbox/net-kourier/pull/1075 instead of here?

PR #1068, opened 2023-06-23
Configure tracing for kourier-gateway (Envoy)

norbjd:
# Changes

- :gift: Configure tracing for kourier-gateway (Envoy)

/kind enhancement

Hello :wave:

It could be very useful to configure tracing at the gateway level (Envoy), for example to follow an external HTTP request from the moment when it reaches the gateway to the knative service itself. Until now, we were only able to configure traces starting from the `activator` service, which is already not bad, but:

- not having tracing at the gateway level misses some requests like 4xx or 5xx errors thrown by Envoy
- adding tracing here also allow to measure the latency between the gateway and the activator service, and can help troubleshooting

For this PR, I have added the tracing config to every listener, so that means we have traces for every request that reaches the gateway, even for internal traffic (Knative Ingress Probe, or Kube probe).

**Demo**

By configuring the `config-kourier` configmap like this:

```yaml
data:
  tracing-enabled: "true"
  tracing-collector-host: "jaeger.tracing.svc.cluster.local"
  tracing-collector-port: "9411"
  tracing-collector-endpoint: "/api/v2/spans"
```

We were able to get a new `kourier-knative` span (before the `activator-service`):

image

Clicking on the trace gives various details like: the host, the http method, some headers, etc.

In the previous example, we were using Grafana Tempo as the collector, but it should work fine with everything supporting Zipkin.
 
**Release Note**

```release-note
Add tracing support at the gateway level
```

**Docs**

WIP, I'll do this once the PR has been reviewed (because things can change in the meantime, based on your comments).


nak3:
Super awesome! Thank you so much @norbjd I will review this next week (though next Monday is public holiday so might be slow to response).

nak3:
Overall looks good to me. One question before moving on the detailed review, how do you think about using `config-tracing` especially [zipkin-endpoint](https://github.com/knative/serving/blob/main/config/core/configmaps/tracing.yaml#L48) and [backnd](https://github.com/knative/serving/blob/main/config/core/configmaps/tracing.yaml#L44C5-L44C12) instead of introducing the new fields in `config-kourier`?

As controller already uses `config-logging` and `config-observability`, `config-tracing`, it makes sense?

nak3:
I see. Then, let's use `config-kourier` like you have already done.
With said, can we use one configuration parameter like?

```
data:
  tracing-collector-endpoint: "jaeger.tracing.svc.cluster.local:9411/api/v2/spans"
```

Then, parse the edpoint and enable/disable via omit or specified.

pkg/config/configmap.go
```diff
@@ -54,6 +58,8 @@ const (
 
 	// enableCryptoMB is the config map for enabling CryptoMB private key provider.
 	enableCryptoMB = "enable-cryptomb"
+
```
nak3, line 61: Coould you add a comment for `TracingCollectorFullEndpoint` ?

pkg/config/configmap.go
```diff
@@ -82,13 +88,50 @@ func NewConfigFromMap(configMap map[string]string) (*Kourier, error) {
...
 		cm.AsBool(enableCryptoMB, &nc.EnableCryptoMB),
+		asTracing(TracingCollectorFullEndpoint, &nc.Tracing),
 	); err != nil {
 		return nil, err
 	}
 
 	return nc, nil
 }
 
+type Tracing struct {
```
nak3, line 99: It would be also great if you could add a comment about `Tracing` struct.

pkg/config/configmap.go
```diff
@@ -82,13 +88,50 @@ func NewConfigFromMap(configMap map[string]string) (*Kourier, error) {
...
+
+func asTracing(collectorFullEndpoint string, tracing *Tracing) cm.ParseFunc {
+	return func(data map[string]string) error {
+		if raw, ok := data[collectorFullEndpoint]; ok && raw != "" {
+			tracing.Enabled = true
+
+			// We add a random scheme to be able to use url.ParseRequestURI.
+			parsedURL, err := url.ParseRequestURI("scheme://" + raw)
+			if err != nil {
+				return fmt.Errorf("\"%s\" is not a valid URL: %w", raw, err)
```
nak3, line 114:
```suggestion
				return fmt.Errorf("%q is not a valid URL: %w", raw, err)
```

nit

pkg/config/configmap.go
```diff
@@ -82,13 +88,50 @@ func NewConfigFromMap(configMap map[string]string) (*Kourier, error) {
...
+			}
+
+			tracing.CollectorHost = parsedURL.Hostname()
+			collectorPortUint64, err := strconv.ParseUint(parsedURL.Port(), 10, 32)
+			if err != nil {
+				return fmt.Errorf("\"%s\" is not a valid port: %w", parsedURL.Port(), err)
```
nak3, line 120:
```suggestion
				return fmt.Errorf("%q is not a valid port: %w", parsedURL.Port(), err)
```

pkg/generator/caches.go
```diff
@@ -221,14 +226,14 @@ func (caches *Caches) deleteTranslatedIngress(ingressName, ingressNamespace stri
 	}
 }
 
-func generateListenersAndRouteConfigs(
+func generateListenersAndRouteConfigsAndClusters(
```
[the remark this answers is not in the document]
nak3: Could you please open the github issue to refactor around here?

test/tracing/tracing_test.go
```diff
@@ -0,0 +1,193 @@
+//go:build e2e
+// +build e2e
+
+/*
+Copyright 2020 The Knative Authors
```
nak3, line 5:
```suggestion
Copyright 2023 The Knative Authors
```

test/e2e-kind.sh
```diff
@@ -173,6 +173,31 @@ kubectl -n "${KOURIER_CONTROL_NAMESPACE}" patch configmap/config-kourier --type
...
+echo ">> Setup Tracing"
+kubectl apply -f test/config/tracing
+kubectl -n tracing wait --timeout=300s --for=condition=Available deployment/jaeger
+export TRACING_COLLECTOR_FULL_ENDPOINT="$(kubectl -n tracing get svc/jaeger -o jsonpath='{.spec.clusterIP}'):9411/api/v2/spans"
+kubectl -n "${KOURIER_CONTROL_NAMESPACE}" patch configmap/config-kourier --type merge -p "{
+  \"data\":{
+    \"tracing-collector-full-endpoint\": \"$TRACING_COLLECTOR_FULL_ENDPOINT\"
+  }
+}"
+kubectl -n "${KOURIER_CONTROL_NAMESPACE}" rollout restart deployment/net-kourier-controller
```
nak3, line 185: The `rollout restart/status` should not be necessary as `config-kourier` change can be applied without restart.

We will remove other places like https://github.com/knative-sandbox/net-kourier/pull/1077

nak3: It is not a new behavior. The controller watches the configmap which is stored in `./pkg/reconciler/ingress/config/store.go`  (`config-network` and `config-kourier`) via  https://github.com/knative-sandbox/net-kourier/blob/c8bcc9ae38e3b77270ebe48b0de8612d9a56b32a/pkg/reconciler/ingress/controller.go#L102-L107

Also IIRC, kourier reloads `conig-logging` and `config-observability` automatically thanks to the implementation of pkg's shared main.

nak3:
/lgtm
/approve

Thank you!

PR #1072, opened 2023-07-07
Revert "Add startup probe for controller (#1063)"

nak3:
This reverts commit 039e48afbca8d73223a59a29c0b5220cb74e4614.

Since https://github.com/knative-sandbox/net-kourier/pull/1066 fixes the root cause, the startup time is much faster even when there are many resources. There is no harm even if startup probe exists but we drop it because any other knative repo does not have it. 

/cc @ReToCode


ReToCode (approved):
/lgtm

PR #1075, opened 2023-07-11
Use informer for startupTranslator

nak3:
This patch uses informer for `startupTranslator` instead of k8s client.

This is alternative of https://github.com/knative-sandbox/net-kourier/pull/1066

### DIFF
--- a/pkg/reconciler/ingress/controller.go
+++ b/pkg/reconciler/ingress/controller.go
@@ -25,6 +25,7 @@ import (
 	"go.uber.org/zap"
 	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/labels"
 	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/apimachinery/pkg/util/sets"
 	v1 "k8s.io/client-go/informers/core/v1"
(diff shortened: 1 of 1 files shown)


pkg/reconciler/ingress/controller.go
```diff
@@ -204,26 +203,24 @@ func NewController(ctx context.Context, cmw configmap.Watcher) *controller.Impl
...
 		logger.Fatalw("Failed to fetch ready ingresses", zap.Error(err))
 	}
 	logger.Infof("Priming the config with %d ingresses", len(ingressesToSync))
 
-	// The startup translator uses clients instead of listeners to correctly list all
-	// resources at startup.
 	startupTranslator := generator.NewIngressTranslator(
 		func(ns, name string) (*corev1.Secret, error) {
-			return kubernetesClient.CoreV1().Secrets(ns).Get(ctx, name, metav1.GetOptions{})
+			return secretInformer.Lister().Secrets(ns).Get(name)
```
ReToCode, line 214: Just to make sure, we are now using the filtered secretInformer instead of the unfiltered kubeclient. It seems that Kourier can have just the secret filtered here: https://github.com/knative-sandbox/net-kourier/blob/main/pkg/reconciler/informerfiltering/util.go#L44. So this should be fine, right?

ReToCode (approved):
Other than that:

/lgtm

/hold for the one comment. Feel free to unhold if this is ok.

pkg/reconciler/ingress/controller.go
nak3: Yes, I think no problem. 

Secrets for internal-encryption has `networking.internal.knative.dev/certificate-uid` label in serving repo and Secrets for DomainMapping BYO should be added `networking.internal.knative.dev/certificate-uid` by users so we just need to take care of the filtered secrets when users enabled the filtering.

PR #1077, opened 2023-07-24
Drop rollout for e2e test which updates config-kourier

nak3:
This patch drops rollout for e2e test which updates config-kourier.

Since we don't need to restart net-kourier-controller when changing `config-kourier`, this patch removes the `rollout` commands.

The `rollout restart` should be added only for env variable change tests such as `*_EXTAUTHZ_` as it restarts controller and we need to wait until new pods are up.

/kind cleanup


ReToCode (approved):
/lgtm

PR #1084, opened 2023-08-02
[WIP] Add filtering label to secret and enable secret informer filtering by default

nak3:
**Changes**

Currently users have to set `ENABLE_SECRET_INFORMER_FILTERING_BY_CERT_UID` to enable secret informer filtering.
It has a big pain that they have to add a label to their every single secret on their DomainMapping's BYO certs.

This patch changes the controller to add the label to the secret specified by DomainMapping.
As a consequence, it can enable `ENABLE_SECRET_INFORMER_FILTERING_BY_CERT_UID` by default.

/kind enhancement

**Release Note**

```release-note
Kourier controller injects labels (`networking.internal.knative.dev/certificate-uid: <secret_name>`) to the secret used by auto-tls or DomainMapping's BYO certs.
It make possible to enables secret informer filtering (enabled via `ENABLE_SECRET_INFORMER_FILTERING_BY_CERT_UID` for now) by defalut.
```

TODO:
- [ ] update unit test


nak3:
@skonto @dprotaso Could you please take a look if this approach is alright? I will add tests if no problem.
It is related to https://github.com/knative-extensions/net-kourier/pull/862 https://github.com/knative-extensions/net-istio/pull/920

pkg/generator/ingress_translator.go
```diff
@@ -89,10 +95,31 @@ func (translator *IngressTranslator) translateIngress(ctx context.Context, ingre
 		}
 
 		secret, err := translator.secretGetter(ingressTLS.SecretNamespace, ingressTLS.SecretName)
-		if err != nil {
+		if apierrors.IsNotFound(err) {
```
skonto, line 98: So we are trying to make the UX easier and we add the label ourselves.
Are there any other secrets that might need the same?

nak3: Other secrets are "static" secret and we have already added the label like https://github.com/knative/serving/blob/main/config/core/300-secret.yaml#L34 by default. So we don't need it.

nak3:
So, is it okay for this approach?

nak3:
Closing due to no feedback.

PR #1108, opened 2023-09-13
Do not use informer for startupTranslator

nak3:
Fix https://github.com/knative-extensions/net-kourier/issues/1106

As previous comment line explains `The startup translator uses clients instead of listeners to correctly list all resources at startup.`, the informer cannot fetch the secret.

This patch reverts to use informer.

**Release Note**

```release-note
Revert the usage of informer during startupTranslator
```

### DIFF
--- a/pkg/reconciler/ingress/controller.go
+++ b/pkg/reconciler/ingress/controller.go
@@ -204,15 +204,17 @@ func NewController(ctx context.Context, cmw configmap.Watcher) *controller.Impl 
 	}
 	logger.Infof("Priming the config with %d ingresses", len(ingressesToSync))
 
+	// The startup translator uses clients instead of listeners to correctly list all
+	// resources at startup.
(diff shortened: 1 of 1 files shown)


nak3:
This only happens on secret because service handles `Not Found` error like https://github.com/knative-extensions/net-kourier/blob/26fa09e9b37ffa22a689044bd4d82c9007d3f26b/pkg/generator/ingress_translator.go#L142-L149

I considered to apply it to endpoint https://github.com/knative-extensions/net-kourier/compare/main...nak3:handle-secret-not-found?expand=1 but it will not make sense for startupTranslator.

nak3:
https://github.com/knative-extensions/net-kourier/issues/1106 can be reproduced if Kingress has a secret - i.e. DomainMaping with BYO or enabling autoTLS.

ReToCode (approved):
/lgtm
/approve

PR #1116, opened 2023-09-19
Use aligned flags and secrets for `system-internal-tls`

ReToCode:
# Changes
- Uses the aligned flags from https://github.com/knative/serving/issues/14368
- Uses the new secrets & SANs from https://github.com/knative/serving/issues/14369

/kind cleanup

Partially https://github.com/knative/serving/issues/14369
Partially https://github.com/knative/serving/issues/14392
Partially https://github.com/knative/serving/issues/14368

**Release Note**

```release-note
net-kourier now uses the new `system-internal-tls` flag to configure encryption of Knative system components.
```

**Docs**
Will be done separately when encryption is working end-to-end


nak3:
/hold

Ah, `Code Style / style / Golang / Lint` should be fixed.

ReToCode:
We also need to create the new certificates with SANs in the script. 
For now we do not use all of them, but I think it is better to have the correct secrets with SANs that we are going to use in Serving, so we can build on top of that when adding the `cluster-local-domain-tls` stuff.

ReToCode:
/unhold
@nak3

PR #1149, opened 2023-10-27
Increase kourier controller CPU limit

ReToCode:
# Changes
- Increases the CPU limits of net-kourier controller to avoid slow startup with ~500 KSVCs

**Release Note**

```release-note
The CPU limit of kourier controller was increased to avoid CPU limitation issues in larger installations
```

### DIFF
--- a/config/300-controller.yaml
+++ b/config/300-controller.yaml
@@ -94,7 +94,7 @@ spec:
               cpu: 200m
               memory: 200Mi
             limits:
-              cpu: 500m
+              cpu: "1"
               memory: 500Mi
       restartPolicy: Always
       serviceAccountName: net-kourier


nak3:
/lgtm

But will we backport this to 1.10 and other branches on upstream? I wonder that we should backport it into our downstream instead of creating the new releases i.e. v1.11.6, v1.10.4 etc. I don't have strong objections but just wanted to confirm.

ReToCode:
> But will we backport this to 1.10 and other branches on upstream? I wonder that we should backport it into our downstream instead of creating the new releases i.e. v1.11.6, v1.10.4 etc. I don't have strong objections but just wanted to confirm.

Valid point, let's do that then, I'll close the PRs that will be opened.

/unhold

ReToCode:
@nak3 I did keep the 1.12 back-port, as this version will be here for a while, and then we can upstream-sync easier.

PR #1156, opened 2023-11-13
Host cluster-local-domain TLS on local listener with SNI

ReToCode:
# Changes
- Add cluster-local https listener when `cluster-local-domain-tls` is enabled and a KSVC exists
- Uses SNI to host multiple certificates for all cluster-local domain Knative Services
- Rename `internal` -> `local` to be consistent with the new encryption flags

/hold requires https://github.com/knative/networking/pull/891 to be merged first.

/kind enhancement

Fixes https://github.com/knative/serving/issues/14218
Partially https://github.com/knative/serving/issues/14624

**Release Note**
```release-note
net-kourier now hosts a TLS listener using SNI when `cluster-local-domain-tls` is enabled. Note: this is an experimental alpha-feature.
```

**Docs**

Will be done once the features is complete

### DIFF
--- a/pkg/generator/caches.go
+++ b/pkg/generator/caches.go
@@ -39,18 +39,17 @@ import (
 	"knative.dev/net-kourier/pkg/config"
 	envoy "knative.dev/net-kourier/pkg/envoy/api"
 	rconfig "knative.dev/net-kourier/pkg/reconciler/ingress/config"
+	"knative.dev/networking/pkg/certificates"
 	"knative.dev/pkg/system"
 )
 
 const (
 	envCertsSecretNamespace    = "CERTS_SECRET_NAMESPACE"
 	envCertsSecretName         = "CERTS_SECRET_NAME"
-	certFieldInSecret          = "tls.crt"
-	keyFieldInSecret           = "tls.key"
 	externalRouteConfigName    = "external_services"
 	externalTLSRouteConfigName = "external_tls_services"
-	internalRouteConfigName    = "internal_services"
-	internalTLSRouteConfigName = "internal_tls_services"
+	localRouteConfigName       = "internal_services"
+	localTLSRouteConfigName    = "internal_tls_services"
 )
 
 // ErrDomainConflict is an error produces when two ingresses have conflicting domains.
@@ -66,7 +65,7 @@ type Caches struct {
 	kubeClient kubeclient.Interface
 }
 
-func NewCaches(ctx context.Context, kubernetesClient kubeclient.Interface, extAuthz bool) (*Caches, error) {
+func NewCaches(kubernetesClient kubeclient.Interface, extAuthz bool) (*Caches, error) {
 	c := &Caches{
 		translatedIngresses: make(map[types.NamespacedName]*translatedIngress),
 		clusters:            newClustersCache(),
@@ -92,7 +91,7 @@ func (caches *Caches) UpdateIngress(ctx context.Context, ingressTranslation *tra
 }
 
 func (caches *Caches) validateIngress(translatedIngress *translatedIngress) error {
-	for _, vhost := range translatedIngress.internalVirtualHosts {
+	for _, vhost := range translatedIngress.localVirtualHosts {
 		if caches.domainsInUse.HasAny(vhost.Domains...) {
 			return ErrDomainConflict
 		}
@@ -106,7 +105,7 @@ func (caches *Caches) addTranslatedIngress(translatedIngress *translatedIngress)
 		return err
 	}
 
-	for _, vhost := range translatedIngress.internalVirtualHosts {
+	for _, vhost := range translatedIngress.localVirtualHosts {
 		caches.domainsInUse.Insert(vhost.Domains...)
 	}
 
@@ -135,17 +134,23 @@ func (caches *Caches) ToEnvoySnapshot(ctx context.Context) (*cache.Snapshot, err
 	defer caches.mu.Unlock()
 
 	localVHosts := make([]*route.VirtualHost, 0, len(caches.translatedIngresses)+1)
+	localTLSVHosts := make([]*route.VirtualHost, 0, len(caches.translatedIngresses)+1)
 	externalVHosts := make([]*route.VirtualHost, 0, len(caches.translatedIngresses))
 	externalTLSVHosts := make([]*route.VirtualHost, 0, len(caches.translatedIngresses))
-	snis := sniMatches{}
+	localSNIs := sniMatches{}
+	externalSNIs := sniMatches{}
 
 	for _, translatedIngress := range caches.translatedIngresses {
-		localVHosts = append(localVHosts, translatedIngress.internalVirtualHosts...)
+		localVHosts = append(localVHosts, translatedIngress.localVirtualHosts...)
+		localTLSVHosts = append(localTLSVHosts, translatedIngress.localTLSVirtualHosts...)
 		externalVHosts = append(externalVHosts, translatedIngress.externalVirtualHosts...)
 		externalTLSVHosts = append(externalTLSVHosts, translatedIngress.externalTLSVirtualHosts...)
 
-		for _, match := range translatedIngress.sniMatches {
-			snis.consume(match)
+		for _, match := range translatedIngress.localSNIMatches {
+			localSNIs.consume(match)
+		}
+		for _, match := range translatedIngress.externalSNIMatches {
+			externalSNIs.consume(match)
 		}
 	}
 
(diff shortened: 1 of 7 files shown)


ReToCode:
PTAL, I'll rebase and remove the `go mod replace` once the networking PR is merged.

/assign @nak3 
/assign @skonto

pkg/generator/ingress_translator.go
```diff
@@ -48,11 +48,13 @@ import (
 
 type translatedIngress struct {
 	name                    types.NamespacedName
-	sniMatches              []*envoy.SNIMatch
+	localSNIMatches         []*envoy.SNIMatch
```
skonto, line 51: Thinking out loud is there a scenario where a ksvc wants to communicate with an egress target via the knative local gateway and the target does not need mutual tls. I was wondering about the use of [auto_sni](https://www.envoyproxy.io/docs/envoy/latest/api-v3/config/core/v3/protocol.proto#envoy-v3-api-field-config-core-v3-upstreamhttpprotocoloptions-auto-sni).

pkg/generator/caches_test.go
```diff
@@ -166,14 +166,14 @@ func TestDeleteIngressInfoWhenDoesNotExist(t *testing.T) {
 	assert.DeepEqual(t, listenersBeforeDelete, listenersAfterDelete, protocmp.Transform())
 }
 
-func TestTLSListenerWithEnvCertsSecret(t *testing.T) {
+func TestExternalTLSListener(t *testing.T) {
```
skonto, line 169: We don't want to support
- TestExternalTLSListenerWithEnvCertsSecrets 
- TestExternalTLSListenerWithInternalCertSecret
Local is only meant to use the config map config?

pkg/generator/ingress_translator.go
```diff
@@ -103,14 +106,26 @@ func (translator *IngressTranslator) translateIngress(ctx context.Context, ingre
...
 			Hosts:            ingressTLS.Hosts,
 			CertSource:       secretRef,
-			CertificateChain: secret.Data[certFieldInSecret],
-			PrivateKey:       secret.Data[keyFieldInSecret]})
+			CertificateChain: secret.Data[certificates.CertName],
+			PrivateKey:       secret.Data[certificates.PrivateKeyName],
+		}
+
+		switch ingressTLS.Visibility {
+		// Also handle empty string as external for reverse-compatibility
```
skonto, line 117: nit:s/reverse/backwards

skonto: I am wondering if we need to verify ingressTLS.Visibility (previously we didn't in this controller)? 
If not this could be just: 
```
if ingressTLS.Visibility == v1alpha1.IngressVisibilityClusterLocal {
  localSNIMatches = append(localSNIMatches, sniMatch)
} else {
  externalSNIMatches = append(externalSNIMatches, sniMatch)
}
```

pkg/generator/caches.go
```diff
@@ -251,31 +260,74 @@ func generateListenersAndRouteConfigsAndClusters(
...
 	listeners = append(listeners, probHTTPListener)
 
-	// Add internal listeners and routes when internal cert secret is specified.
-	if cfg.Kourier.ClusterCertSecret != "" {
-		internalTLSRouteConfig := envoy.NewRouteConfig(internalTLSRouteConfigName, clusterLocalVirtualHosts)
-		internalTLSManager := envoy.NewHTTPConnectionManager(internalTLSRouteConfig.Name, cfg.Kourier)
+	// Configure TLS listener for cluster-local traffic.
+	// If there's at least one ingress that contains the TLS field, that takes precedence.
+	// If there is not, TLS will be configured using a single cert for all the services when the certificate is configured.
+	if len(localSNIMatches) > 0 {
```
skonto, line 266: Could we refactor part of this behind a func for both local and external names, it seems there is some repetition?

pkg/generator/caches.go
```diff
@@ -135,17 +134,23 @@ func (caches *Caches) ToEnvoySnapshot(ctx context.Context) (*cache.Snapshot, err
 	defer caches.mu.Unlock()
 
 	localVHosts := make([]*route.VirtualHost, 0, len(caches.translatedIngresses)+1)
+	localTLSVHosts := make([]*route.VirtualHost, 0, len(caches.translatedIngresses)+1)
```
skonto, line 137: We could leave it as a nil slice here in case there are no local tls ingresses (thus no need to have empty slices in tests) but no strong preference as there is no performance diff anyway.

skonto:
LGTM in general waiting for @nak3 before stamping.

pkg/generator/ingress_translator.go
ReToCode: Hm sounds to me more like a Service-Mesh case. I don't think we ever want to go through kourier for egress traffic, right?

pkg/generator/caches_test.go
ReToCode: I don't think I get your question.

The ExternalTLS Listener is for cluster-external domains and the cluster-local one is for the cluster-local domains. Both can have the SNI option (has already been there for the external one, this PR adds it for cluster-local) when a specific cert is there for a specific domain, and a general certificate for all other traffic (that is already there for both listeners).

pkg/generator/ingress_translator.go
ReToCode: I think that would also work, but could have some bad runtime behaviour if we ever add another option.

pkg/generator/caches.go
ReToCode: 👍 I'll take a look

pkg/generator/caches.go
ReToCode: Good point, I just went with the same approach as with the other initialisations.

pkg/generator/caches.go
ReToCode: Hm they are similar but not equal, differences:
- Route config is named differently on cluster-local between the two ifs, but not on external
- Filter chains methods calls are different (`newLocalEnvoyListenerWithOneCertFilterChain`)
- One cert in external is configured with env variables, one cert in cluster-local is configured using a secret from the CM

So that extracted method would have quite some params and also a passed in function to lookup the filter chains based on different methods. Not sure if we benefit enough to refactor.

pkg/generator/caches_test.go
skonto: Yes but the tests were renamed and the source of the cert can be either env or some cm config. What do we support for internal and external cases?

pkg/generator/caches.go
skonto: The func is too long tbh but we can always improve later.

pkg/reconciler/ingress/lister.go
```diff
@@ -63,30 +63,40 @@ func (l *gatewayPodTargetLister) ListProbeTargets(ctx context.Context, ing *v1al
...
-			target = status.ProbeTarget{
-				PodIPs:  ips,
-				PodPort: strconv.Itoa(int(config.HTTPPortInternal)),
-				URLs:    domainsToURL(domains, scheme),
-			}
+		target := status.ProbeTarget{
+			PodIPs: ips,
+		}
+
+		switch {
```
nak3, line 84: You made this so much clean and readable :heart_eyes_cat:

nak3 (commented):
Do you have a plan to add conformance or e2e test for this cluster local domain TLS in networking or net-* repos like external domain for `test/conformance/ingress/tls.go`?

pkg/generator/caches_test.go
ReToCode: I felt the naming did not really reflect what they test. Both cases can have a default cert for all domains and specific ones for specified domains, I think we do cover all of them for both cases with:

```
func TestExternalTLSListener(t *testing.T) {

// sets the default cert
t.Setenv(envCertsSecretNamespace, "certns")
t.Setenv(envCertsSecretName, "secretname")

// Test with only default domains
t.Run("without SNI matches", func(t *testing.T) {}

// Test with one added specific domain
t.Run("with a single SNI match", func(t *testing.T) {}

// Test with two added specific domains
t.Run("with multiple SNI matches", func(t *testing.T) {}
}
```

pkg/reconciler/ingress/lister.go
ReToCode: Thanks 🙏

ReToCode:
> Do you have a plan to add conformance or e2e test for this cluster local domain TLS in networking or net-* repos like external domain for test/conformance/ingress/tls.go?

Yes this is the plan in https://github.com/knative/serving/issues/13855. I'm not sure yet how they can look like exactly. I think we should finish the implementation end to end for one ingress then add conformance tests. Also Serving will get additional end-2-end tests [draft here](https://github.com/knative/serving/pull/14610/files#diff-b640743f3754af98db303a027c84a08a54176c4a467fb77028587518f87630d4).

ReToCode:
/unhold

Networking PR is merged: https://github.com/knative/networking/pull/891, conformance tests is tracked in https://github.com/knative/serving/issues/13855. So this is ready to be reviewed again @nak3 @skonto

skonto:
Tested in https://github.com/knative/serving/pull/14703 stamping.

PR #1157, opened 2023-11-14
Increase Kourier gateway CPU limit

nak3:
Increases the CPU limits of kourier gateway instead of controller to avoid readliness probe failure ~500 KSVCs.

**Release Note**

```release-note
The CPU limit of kourier Gateway limit is now increased to avoid CPU limitation issues in larger installations.
```

/cc @ReToCode @skonto


ReToCode (approved):
/lgtm
/approve

Issue #941, opened 2022-10-19
Kourier at large scale


daraghlowe:
**What's the issue?**
We have started testing Kourier at large scale to see if deployment times are better than Istio(time for a KSVC to become ready to serve traffic). Deploy times are good with Kourier and it consistently takes less than 10 seconds for a newly added KSVC to become ready all the way up to 2000 KSVC.

However, if you delete a KSVC and then you try to add a new KSVC, times are much slower and even with only 500 KSVC on the cluster it takes several minutes before the new KSVC is ready.

Looking at the logs in the net-kourier-controller, you can see that it starts reconciling all of the Ingress on the cluster when you delete a KSVC and presumably this needs to finish before the new ingress can be created for our new KSVC.

**Why is this a problem?**
This leads to inconsistent deploy times for our workloads which creates an inconsistent user experience as sometimes its really quick and other times it could takes minutes to become ready.

**Results**
Here are the times it took for a single KSVC to become ready right after I deleted a different single KVSC alongside the number of KSVC that were on the cluster.

<img width="203" alt="image" src="https://user-images.githubusercontent.com/22292234/196718079-8bf6724e-7e5c-4a8f-8322-67115224d864.png">

**Why are we doing this?**
We are running a cluster with Knative and Istio with 1500 KSVC and have started to run a problem with the time it's taking before new KSVC we add become ready (the ingress). 

We opened an issue for this here: https://github.com/knative/serving/issues/13247

dprotaso:
I'm going to close this out - as PR https://github.com/knative-extensions/net-kourier/pull/1159 seems to have resolved the issue.

Testing it manually now - I've created 2K Knative Services and deleted the first ten. Subsequent creations take roughly the same amount of time in a kind cluster ~10s

PR #1159, opened 2023-11-20
Fixed reconcile running on all KIngress after ksvc deletion

ryutoyasugi:
## Why
I found Reconcile running on all KIngress after ksvc was removed.
The same problem has been reported. #941 

## Changes
Fixed conditions under which `resyncConflicts` are executed.
https://github.com/knative-extensions/net-kourier/blob/cbd0d74/pkg/reconciler/ingress/ingress.go#L135

## Related PR
#502, #486


skonto:
I tested it locally with conflicting dms, stamping it.
/lgtm

PR #1171, opened 2023-12-06
Add upstream TLS trust from CM bundles

ReToCode:
# Changes
- Adding support to trust multiple CAs (when a label is present) for upstream TLS connections
- Checks if CAs are parsable before adding them to envoy
- Additional to reading the `ca.crt` in activators cert secret, net-kourier now also adds additional trust bundles to the trust pool
- Dropped the fallback implementation of reading service port names. With the above config read fix, this is no longer necessary
- Tested the full set with CA rotation [here](https://github.com/ReToCode/knative-encryption/blob/main/8-trust-sources/POC.md#full-test-set-with-rotation)

Partially https://github.com/knative/serving/issues/14609

**Release Note**

```release-note
net-kourier now allows to add additional CA trust bundles in ConfigMaps to verify upstream TLS connections
```

**Docs**

Will be added once the full encryption feature is completed.


ReToCode:
@skonto @dprotaso what do you think of that approach for https://github.com/knative/serving/issues/14609. 

With this approach, this is not tied to trust-manager. ConfigMaps seem to be a very common interface for CA bundles, see https://github.com/ReToCode/knative-encryption/tree/main/8-trust-sources. 

We can also extend this to https://github.com/knative/serving/issues/14609#issuecomment-1825697583 once this has landed. But I'd vote for reading it using the K8s clients instead of mounting it to the filesystem, as this solves the reloading concern.

skonto:
> ConfigMaps seem to be a very common interface for CA bundles,

In general I see secrets being used as well. For trust manager it seems cm are first option. I don't have strong preference.

pkg/generator/ingress_translator.go
```diff
@@ -194,22 +208,10 @@ func (translator *IngressTranslator) translateIngress(ctx context.Context, ingre
...
-				// running when booting the controller up and prefilling the config before making it
-				// ready.
-				//
-				// TODO:
-				// Drop this configmap check - issues/968
-				// We could determine whether system-internal-tls is enabled or disabled via the flag only,
-				// but all conformance tests need to be updated to have the port name so we check the configmap as well.
-				//
-				// TODO: Or fetch configmap before the loop as per https://github.com/knative-sandbox/net-kourier/pull/959#discussion_r1048441513
-				cfg := config.FromContextOrDefaults(ctx)
```
skonto, line 207: We don't need `config.FromContextOrDefaults(ctx)` any more?

pkg/generator/ingress_translator_test.go
```diff
@@ -1703,6 +2009,42 @@ var (
...
+	validCAConfigmap = &corev1.ConfigMap{
+		ObjectMeta: metav1.ObjectMeta{
+			Namespace: "knative-testing",
+			Name:      "valid-ca",
+			Labels: map[string]string{
+				informerfiltering.KnativeCABundleLabelKey: "true",
+			},
+		},
+		Data: map[string]string{
+			certificates.CaCertName: string(cert),
```
skonto, line 2021: Maybe we should have multiple cas to test against?

ReToCode:
> In general I see secrets being used as well. For trust manager it seems cm are first option. I don't have strong preference.

Hm where do you mean? I don't think we have a current solution (other than `ca.crt`) that would put the bundles in a Secret. But we can always extend if necessary.

pkg/generator/ingress_translator.go
```diff
@@ -106,6 +113,17 @@ func (translator *IngressTranslator) translateIngress(ctx context.Context, ingre
 	externalTLSHosts := make([]*route.VirtualHost, 0, len(ingress.Spec.Rules))
 	clusters := make([]*v3.Cluster, 0, len(ingress.Spec.Rules))
 
+	cfg := config.FromContext(ctx)
```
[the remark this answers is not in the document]
ReToCode: No, I added [this](https://github.com/knative-extensions/net-kourier/pull/1171/files#diff-1c1c1dc4a9824c0156196bdf2594574cefc8099a0a67e2e392044468bc6f6a5dR341) to the controller, so we have the CMs now properly on the context, before calling the ingress translator.

pkg/generator/ingress_translator_test.go
ReToCode: 👍  I'll do that.

pkg/generator/ingress_translator.go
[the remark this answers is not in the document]
skonto: Is it possible that the cm will not be deployed yet when we reach the kubeclient call. I was reading again my comment [here](https://github.com/knative-extensions/net-kourier/pull/959#discussion_r1048441513), so sharedMain has a cm watcher that will wait for cms to exist and even if we try to [fetch certain cms earlier and fail (since they dont exist)](https://github.com/knative/pkg/blob/main/injection/sharedmain/main.go#L408-L410) we will get them finally due to the cmw.Start() call.

ReToCode:
> I was referring to the docs: https://cert-manager.io/docs/trust/trust-manager

That is just for sources when I'm reading it right. The result to mount in workloads is always a Configmap?

pkg/generator/ingress_translator.go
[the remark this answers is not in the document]
skonto: I guess the idea was to be eventually be consistent. However, I am in favor of fixing it properly. I would try the wait approach so we dont have to think about corner cases.

ReToCode:
@skonto this is ready now, please take another look.

/unhold

pkg/reconciler/ingress/controller.go
```diff
@@ -81,6 +85,7 @@ func NewController(ctx context.Context, cmw configmap.Watcher) *controller.Impl
 	serviceInformer := serviceinformer.Get(ctx)
 	podInformer := podinformer.Get(ctx)
 	secretInformer := getSecretInformer(ctx)
+	nsConfigmapInformer := nsconfigmapinformer.Get(ctx) // this is filtered to SYSTEM_NAMESPACE
```
skonto, line 88: Good to have that comment there.

pkg/generator/ingress_translator.go
```diff
@@ -144,10 +161,9 @@ func (translator *IngressTranslator) translateIngress(ctx context.Context, ingre
 				// Match the ingress' port with a port on the Service to find the target.
 				// Also find out if the target supports HTTP2.
 				var (
-					externalPort  = int32(80)
-					targetPort    = int32(80)
-					http2         = false
-					httpsPortUsed = false
```
skonto, line 150: Why we had this one in the first place?

pkg/reconciler/ingress/controller.go
```diff
@@ -355,6 +384,16 @@ func getSecretInformer(ctx context.Context) v1.SecretInformer {
 	return secretfilteredinformer.Get(ctx, untyped.([]string)[0])
 }
 
+func getLabelSelector(label string) (labels.Selector, error) {
+	selector := labels.NewSelector()
+	req, err := labels.NewRequirement(label, selection.Exists, make([]string, 0))
```
skonto, line 389: I think we can avoid the allocation:
```labels.NewRequirement(label, selection.Exists,nill)``` correct?

File: pkg/reconciler/ingress/controller.go [hunk 7]
```diff
@@ -183,6 +188,13 @@ func NewController(ctx context.Context, cmw configmap.Watcher) *controller.Impl
 		func(ns, name string) (*corev1.Secret, error) {
 			return secretInformer.Lister().Secrets(ns).Get(name)
 		},
+		func(label string) ([]*corev1.ConfigMap, error) {
+			selector, err := getLabelSelector(label)
+			if err != nil {
+				return nil, err
+			}
+			return nsConfigmapInformer.Lister().ConfigMaps(system.Namespace()).List(selector)
```
skonto, line 196: Simplify to:

```
 nsConfigmapInformer.Lister().List(selector)
```
since we have a cm informer based on ns.

pkg/generator/ingress_translator.go
```diff
@@ -466,3 +517,24 @@ func domainsForRule(rule v1alpha1.IngressRule) []string {
 	}
 	return domains
 }
+
+func checkCertBundle(certs []byte) error {
```
skonto, line 521: It seems that this error is not being caught by decode:

```
var rsaSecretCertPEM = `-----BEGIN CERTIFICATE-----
-----BEGIN CERTIFICATE-----
MIIB0zCCAX2gAwIBAgIJAI/M7BYjwB+uMA0GCSqGSIb3DQEBBQUAMEUxCzAJBgNV
BAYTAkFVMRMwEQYDVQQIDApTb21lLVN0YXRlMSEwHwYDVQQKDBhJbnRlcm5ldCBX
aWRnaXRzIFB0eSBMdGQwHhcNMTIwOTEyMjE1MjAyWhcNMTUwOTEyMjE1MjAyWjBF
MQswCQYDVQQGEwJBVTETMBEGA1UECAwKU29tZS1TdGF0ZTEhMB8GA1UECgwYSW50
ZXJuZXQgV2lkZ2l0cyBQdHkgTHRkMFwwDQYJKoZIhvcNAQEBBQADSwAwSAJBANLJ
hPHhITqQbPklG3ibCVxwGMRfp/v4XqhfdQHdcVfHap6NQ5Wok/4xIA+ui35/MmNa
rtNuC+BdZ1tMuVCPFZcCAwEAAaNQME4wHQYDVR0OBBYEFJvKs8RfJaXTH08W+SGv
zQyKn0H8MB8GA1UdIwQYMBaAFJvKs8RfJaXTH08W+SGvzQyKn0H8MAwGA1UdEwQF
MAMBAf8wDQYJKoZIhvcNAQEFBQADQQBJlffJHybjDGxRMqaRmDhX0+6v02TUKZsW
r5QuVbpQhH6u+0UgcW0jp9QwpxoPTLTWGXEWBBBurxFwiCBhkQ+V
-----END CERTIFICATE-----
`
````

Comment (skonto) [reply pkg/generator/ingress_translator.go:521]: I played around with decode a bit.
The way it is written:
```
func Decode(data []byte) (p *Block, rest []byte) {
	// pemStart begins with a newline. However, at the very beginning of
	// the byte array, we'll accept the start string without it.
	rest = data
	for {
		if bytes.HasPrefix(rest, pemStart[1:]) {
			rest = rest[len(pemStart)-1:]
		} else if _, after, ok := bytes.Cut(rest, pemStart); ok {
			rest = after
		} else {
			return nil, data
		}

		var typeLine []byte
		typeLine, rest = getLine(rest)
		if !bytes.HasSuffix(typeLine, pemEndOfLine) {
			continue
		}
		typeLine = typeLine[0 : len(typeLine)-len(pemEndOfLine)]

		p = &Block{
			Headers: make(map[string]string),
			Type:    string(typeLine),
		}
...
```
It seems it is always going to consume the input.
I am wondering if there is a scenario where block== nil but we have len(rest) >0.

File: pkg/generator/ingress_translator.go [hunk 9]
```diff
@@ -466,3 +517,24 @@ func domainsForRule(rule v1alpha1.IngressRule) []string {
...
+	for block, rest := pem.Decode(certs); block != nil || len(rest) > 0; block, rest = pem.Decode(rest) {
+		if block != nil {
+			switch block.Type {
+			case "CERTIFICATE":
+				_, err := x509.ParseCertificate(block.Bytes)
```
skonto, line 526: Are we planning to do any verification at some point like with this function: https://github.com/istio/istio/blob/master/security/pkg/pki/util/keycertbundle.go#L284 or as soon as we get the cert bundle we assume an external entity is responsible to create valid certs?

pkg/generator/ingress_translator.go
ReToCode: AFAIK: I think because of the config.FromContextOrDefault(). In the first initialization, the default value was encryption disabled. So the service port was used to find out, that is was an encrypted service. Otherwise envoy would have been configured with http until the CMW fired and it reconfigured it again with https.

pkg/generator/ingress_translator.go
ReToCode: Yes I had one case with that, that's why I added

`|| len(rest) > 0`, this way we error out in `checkCertBundle` when there is still a rest that is not consumed.

ReToCode: Yeah I think it parses from comment to comment, so having multiple `-----BEGIN CERTIFICATE-----` is just ignored, as long as the cert within is valid.

pkg/generator/ingress_translator.go
ReToCode: Yes, we give the CA to envoy, envoy will then use it to verify the upstream TLS server cert.

skonto: Yeah just wondering if Istio is doing a pre-validation tep.

pkg/generator/ingress_translator_test.go
```diff
@@ -1273,7 +1282,283 @@ func TestIngressTranslatorWithUpstreamTLS(t *testing.T) {
...
+				localSNIMatches:    []*envoy.SNIMatch{},
+				clusters: []*v3.Cluster{
+					envoy.NewCluster(
+						"servicens/servicename",
+						5*time.Second,
+						lbHTTPSEndpoints,
+						false,
+						&envoycorev3.TransportSocket{
+							Name:       wellknown.TransportSocketTls,
+							ConfigType: typedConfig(false, nil),
```
skonto, line 1491: Passing nil here (also tested with a missing cm) means that we end up having no validation since this is passed to https://github.com/knative-extensions/net-kourier/blob/2a72a55e82ed530aca30a81286ad1b7e6f321761/pkg/generator/ingress_translator_test.go#L2078.
Check comments here too: https://github.com/knative-extensions/net-kourier/blob/2a72a55e82ed530aca30a81286ad1b7e6f321761/vendor/github.com/envoyproxy/go-control-plane/envoy/extensions/transport_sockets/tls/v3/common.pb.go#L816-L835
Would that allow tls communication without validation or connection is failing?
Also do you think that this affects us https://github.com/envoyproxy/envoy/issues/19326 downstream? :thinking:

skonto: What about this part:

" Note 
 // that if a CRL is provided for any certificate authority in a trust chain, a CRL must be 
 // provided for all certificate authorities in that chain. Failure to do so will result in 
 // verification failure for both revoked and unrevoked certificates from that chain."

skonto:
/lgtm
@ReToCode feel free to unhold if you are ready.

PR #1172, opened 2023-12-12
hold startup until ConfigMaps are ready or die

ReToCode:
# Changes
- Follow-up for https://github.com/knative-extensions/net-kourier/pull/959#discussion_r1048441513
- Wait with startupTranslation until Configmaps are found and valid or die
- Later changes to the configmaps are handled through the `configmap.Watcher`
- If configmaps are not present within 60 seconds, the pod will restart with:

```
2023/12/12 14:31:07 Timed out waiting for configuration in ConfigMaps: failed to fetch kourier config: configmaps "config-kourier" not found
Exiting.
```

/assign @skonto

### DIFF
--- a/pkg/reconciler/ingress/controller.go
+++ b/pkg/reconciler/ingress/controller.go
@@ -18,7 +18,10 @@ package ingress
 
 import (
 	"context"
+	"fmt"
+	"log"
 	"strings"
+	"time"
 
 	v3 "github.com/envoyproxy/go-control-plane/envoy/service/discovery/v3"
 	xds "github.com/envoyproxy/go-control-plane/pkg/server/v3"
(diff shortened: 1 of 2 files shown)


pkg/reconciler/ingress/controller.go
```diff
@@ -345,3 +354,42 @@ func getSecretInformer(ctx context.Context) v1.SecretInformer {
 	untyped := ctx.Value(filteredFactory.LabelKey{}) // This should always be not nil and have exactly one selector
 	return secretfilteredinformer.Get(ctx, untyped.([]string)[0])
 }
+
+func ensureCtxWithConfigOrDie(ctx context.Context) context.Context {
+	var err error
+	var cfg *store.Config
+	if pollErr := wait.PollUntilContextTimeout(ctx, time.Second, 60*time.Second, true, func(context.Context) (bool, error) {
```
skonto, line 361: Do we want to make the timeout configurable just in case someone has a slow env?

ReToCode: Hm I don't think thats necessary. It is only until all yamls are applied when installing (from kubectl or operator). I don't think that should ever be longer than 60 seconds, and if so, it would just restart the pod and you have another 60s.

skonto: I don't know about what value is good enough tbh, 1m might be enough. 
Restarting is an option but probably you don't want people to have to do that in case they have a big slow cluster (api server under pressure), just in case.
I mean that if they know they have a slow cluster they can configure it high enough (eg. via env var) and forget about it, otherwise they will have to monitor it and restart it.

PR #1199, opened 2024-02-01
fix: Add metric port to pods and service

arsenetar:
# Changes

This allows the usage of Prometheus PodMonitors and/or ServiceMonitors to configure scraping of metrics.  This is often preferred over the annotations on the pods.  This is also more inline with the rest of Knative.

- Add metric port definition to both gateway pods and controller pods.
- Add metric port to controller service

NOTE: Did not add metric port to gateway services as those often are exposed and exposing the metrics via those services is probably not desired in most cases.

/kind enhancement

**Release Note**

```release-note
Add support for Prometheus ServiceMonitor/PodMonitor
```

**Docs**

```docs

```

Additionally ServiceMonitors and PodMonitors should probably be provided somewhere, maybe over in https://github.com/knative-extensions/monitoring with the rest for base Knative?

### DIFF
--- a/config/300-controller.yaml
+++ b/config/300-controller.yaml
@@ -70,6 +70,9 @@ spec:
           - name: http2-xds
             containerPort: 18000
             protocol: TCP
+          - name: metrics
+            containerPort: 9090
+            protocol: TCP
           readinessProbe:
             grpc:
               port: 18000
@@ -115,6 +118,10 @@ spec:
       port: 18000
       protocol: TCP
       targetPort: 18000
+    - name: http-metrics
+      port: 9090
+      protocol: TCP
+      targetPort: 9090
   selector:
     app: net-kourier-controller
   type: ClusterIP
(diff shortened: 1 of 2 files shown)


arsenetar:
@ReToCode @skonto Is there any feedback on this?

ReToCode (approved):
/lgtm
/approve

/hold for @skonto

ReToCode:
Thanks, I think this makes sense.

> Additionally ServiceMonitors and PodMonitors should probably be provided somewhere, maybe over in https://github.com/knative-extensions/monitoring with the rest for base Knative?

I think that's a good idea. Any input @skonto?

skonto:
Hi all sorry for the late reply, I will take a look. One concern with all metric ports is what info they leak see for example: https://github.com/knative/serving/issues/8959.

arsenetar:
@skonto I don't disagree with the potential for exposed data. If we don't add the port to the controller service then this has no impact vs existing.  For the pods they already have to old annotations for Prometheus, the change to those only allows the pods to be used with the PodMonitors as well which is the more common approach.  It doesn't change the potential exposure at the pod level.

PR #1203, opened 2024-02-02
Gracefully drain connections when stopping the gateway

### LINKED ISSUE #1118: Kourier does not gracefully shut down
when scaling down or updating the `3scale-kourier-gateway` deployment, we see a spike in failed requests, it seems like envoy does not wait for in-flight requests to finish before exiting

norbjd:
# Changes

- :bug: fix gateway preStop hook
- :gift: gracefully drain connections when stopping the gateway

/kind enhancement

This PR supersedes #1200 (click for more details, but TL;DR: today's `preStop` hook doesn't work, because `curl` is not installed in the gateway image).

When the gateway stops, we want to gracefully drain existing connections to prevent users with in-flight requests getting rejected with 5xx errors. To do so, we just need to drain the listeners using [Envoy admin interface and `/drain_listeners?graceful` endpoint](https://www.envoyproxy.io/docs/envoy/v1.26.0/operations/admin#operations-admin-interface-drain).

The `/drain_listeners?graceful` endpoint starts the graceful drain process, but returns immediately. The drain process is configured with `--drain-time-s $(DRAIN_TIME_SECONDS) --drain-strategy immediate`, so the drain will be finished after `DRAIN_TIME_SECONDS`. All connections that couldn't finish after `DRAIN_TIME_SECONDS` will be automatically closed, so this can be configured through the environment variable. People might need to also update `terminationGracePeriodSeconds` depending on their use case (default: 30s).

The reason why I've used `perl` is 1. `perl` is available inside Envoy images and 2. it provides a nice way to connect to Unix socket `/tmp/envoy.admin`. There is also a solution using plain `bash`: https://github.com/knative-extensions/net-kourier/compare/main...norbjd:5be667d697ba34c65946ed42bf4b9533cc27b4c5, but this requires extra changes.

Fixes #1118.

**Release Note**

```release-note
Gracefully drain connections when stopping the gateway
```

### DIFF
--- a/config/300-gateway.yaml
+++ b/config/300-gateway.yaml
@@ -93,7 +93,22 @@ spec:
 
                   exec 3<>/dev/tcp/localhost/9000
                   echo -e "POST /healthcheck/fail HTTP/1.1\nHost: localhost\nConnection: close\n\n" >&3
-                  sleep 15
+
+                  active_connections=1
+
+                  while [ "$active_connections" != "0" ]
+                  do
+                      sleep 1
+
+                      exec 3<>/dev/tcp/localhost/9000
+                      # get downstream_cx_active (active requests) for all "public" listeners (8080, 8081, 8443)
+                      echo -e "GET /stats?filter=listener.0.0.0.0_%288080%7C8081%7C8443%29.downstream_cx_active&format=text HTTP/1.1\nHost: localhost\nConnection: close\n\n" >&3
+
+                      active_connections=$(cat <&3 | grep 'downstream_cx_active' | awk 'BEGIN{active=0} {active+=$NF} END{print active}')
+                      echo "Active connections: $active_connections"
+                  done
+
+                  echo "No more active connections!"
           readinessProbe:
             httpGet:
               httpHeaders:
@@ -123,6 +138,7 @@ spec:
             limits:
               cpu: "1"
               memory: 500Mi
+      terminationGracePeriodSeconds: 120
       volumes:
         - name: config-volume
           configMap:
--- a/config/300-gateway.yaml
+++ b/config/300-gateway.yaml
@@ -91,24 +91,40 @@ spec:
                 - |-
                   set -e
 
-                  exec 3<>/dev/tcp/localhost/9000
-                  echo -e "POST /healthcheck/fail HTTP/1.1\nHost: localhost\nConnection: close\n\n" >&3
+                  # helper method to do a HTTP call to the admin interface
+                  function http_admin_request() {
+                    method="$1"
+                    path="$2"
 
+                    {
+                      echo -e >&3 "$method $path HTTP/1.1\nHost: localhost\nConnection: close\n\n"
+                      cat <&3
+                    } 3<>/dev/tcp/localhost/9000
+                  }
+
+                  function log() {
+                    echo "[$(date --iso-8601=seconds --utc)] $1" > /proc/1/fd/1
+                  }
+
+                  log "Start draining process"
+
+                  # first, run the drain process
+                  http_admin_request POST "/drain_listeners?graceful&skip_exit"
+
+                  # then, wait until there is no more active connections
                   active_connections=1
-
                   while [ "$active_connections" != "0" ]
                   do
                       sleep 1
 
-                      exec 3<>/dev/tcp/localhost/9000
-                      # get downstream_cx_active (active requests) for all "public" listeners (8080, 8081, 8443)
-                      echo -e "GET /stats?filter=listener.0.0.0.0_%288080%7C8081%7C8443%29.downstream_cx_active&format=text HTTP/1.1\nHost: localhost\nConnection: close\n\n" >&3
-
-                      active_connections=$(cat <&3 | grep 'downstream_cx_active' | awk 'BEGIN{active=0} {active+=$NF} END{print active}')
-                      echo "Active connections: $active_connections"
+                      # we use /stats endpoint to find out the number of active connections to listeners, excluding the admin one
+                      active_connections=$(http_admin_request GET "/stats?filter=listener.0.0.0.0_%5B%5E9%5D%5Cd%2B.downstream_cx_active&format=text" | \
+                        grep 'downstream_cx_active' | \
+                        awk 'BEGIN{active=0} {active+=$NF} END{print active}')
+                      log "Number of active connections: $active_connections"
                   done
 
-                  echo "No more active connections!"
+                  log "No more active connections, safely exiting"
           readinessProbe:
             httpGet:
               httpHeaders:
@@ -138,7 +154,7 @@ spec:
             limits:
               cpu: "1"
               memory: 500Mi
-      terminationGracePeriodSeconds: 120
+      terminationGracePeriodSeconds: 300
       volumes:
         - name: config-volume
           configMap:
--- a/config/300-gateway.yaml
+++ b/config/300-gateway.yaml
@@ -48,8 +48,16 @@ spec:
             - --base-id 1
             - -c /tmp/config/envoy-bootstrap.yaml
             - --log-level info
+            - --drain-time-s $(DRAIN_TIME_SECONDS)
+            - --drain-strategy immediate
+            - --parent-shutdown-time-s $(PARENT_SHUTDOWN_TIME_SECONDS)
           command:
             - /usr/local/bin/envoy
+          env:
+            - name: DRAIN_TIME_SECONDS
+              value: "60"
+            - name: PARENT_SHUTDOWN_TIME_SECONDS
+              value: "900"
           image: docker.io/envoyproxy/envoy:v1.26-latest
           name: kourier-gateway
           ports:
@@ -91,40 +99,12 @@ spec:
                 - |-
                   set -e
 
-                  # helper method to do a HTTP call to the admin interface
-                  function http_admin_request() {
-                    method="$1"
-                    path="$2"
-
-                    {
-                      echo -e >&3 "$method $path HTTP/1.1\nHost: localhost\nConnection: close\n\n"
+                  {
+                      echo -e >&3 "POST /drain_listeners?graceful HTTP/1.1\nHost: localhost\nConnection: close\n\n"
                       cat <&3
-                    } 3<>/dev/tcp/localhost/9000
-                  }
+                  } 3<>/dev/tcp/localhost/9000
 
-                  function log() {
-                    echo "[$(date --iso-8601=seconds --utc)] $1" > /proc/1/fd/1
-                  }
-
-                  log "Start draining process"
-
-                  # first, run the drain process
-                  http_admin_request POST "/drain_listeners?graceful&skip_exit"
-
-                  # then, wait until there is no more active connections
-                  active_connections=1
-                  while [ "$active_connections" != "0" ]
-                  do
-                      sleep 1
-
-                      # we use /stats endpoint to find out the number of active connections to listeners, excluding the admin one
-                      active_connections=$(http_admin_request GET "/stats?filter=listener.0.0.0.0_%5B%5E9%5D%5Cd%2B.downstream_cx_active&format=text" | \
-                        grep 'downstream_cx_active' | \
-                        awk 'BEGIN{active=0} {active+=$NF} END{print active}')
-                      log "Number of active connections: $active_connections"
-                  done
-
-                  log "No more active connections, safely exiting"
+                  sleep $DRAIN_TIME_SECONDS
           readinessProbe:
             httpGet:
               httpHeaders:
@@ -154,7 +134,7 @@ spec:
             limits:
               cpu: "1"
               memory: 500Mi
-      terminationGracePeriodSeconds: 300
+      terminationGracePeriodSeconds: 900
       volumes:
         - name: config-volume
           configMap:
--- a/config/300-gateway.yaml
+++ b/config/300-gateway.yaml
@@ -91,17 +91,26 @@ spec:
             preStop:
               exec:
                 command:
-                - /bin/bash
-                - -c
+                - perl
+                - -e
                 - |-
-                  set -e
+                  use strict;
+                  use warnings;
+                  use IO::Socket qw(AF_UNIX SOCK_STREAM SHUT_WR);
 
-                  {
-                      echo -e >&3 "POST /drain_listeners?graceful HTTP/1.1\nHost: localhost\nConnection: close\n\n"
-                      cat <&3
-                  } 3<>/dev/tcp/localhost/9000
+                  my $client = IO::Socket->new(
+                    Domain => IO::Socket::AF_UNIX,
+                    Type   => IO::Socket::SOCK_STREAM,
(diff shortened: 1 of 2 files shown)


config/300-gateway.yaml
```diff
@@ -48,8 +48,13 @@ spec:
             - --base-id 1
             - -c /tmp/config/envoy-bootstrap.yaml
             - --log-level info
+            - --drain-time-s $(DRAIN_TIME_SECONDS)
```
skonto, line 51: Envoy in Istio uses the same idea https://github.com/istio/istio/blob/master/pkg/envoy/proxy.go#L128 the question is what is the best option compared to the current one.


In envoy docs I read that two of the cases where draining happens is when:

The server begins the graceful drain sequence via the [drain_listeners?graceful](https://www.envoyproxy.io/docs/envoy/latest/operations/admin#operations-admin-interface-drain) admin endpoint.

The server has been manually health check failed via the [healthcheck/fail](https://www.envoyproxy.io/docs/envoy/latest/operations/admin#operations-admin-interface-healthcheck-fail) admin endpoint. See the [health check filter](https://www.envoyproxy.io/docs/envoy/latest/intro/arch_overview/upstream/health_checking#arch-overview-health-checking-filter) architecture overview for more information.

So I am wondering if current approach with posting to `http://localhost/healthcheck/fail` would work if we make the sleep time configurable. We probably need to add a test for this with some reproducer if possible.

config/300-gateway.yaml
```diff
@@ -85,7 +90,27 @@ spec:
           lifecycle:
             preStop:
               exec:
-                command: ["/bin/sh","-c","curl -X POST --unix /tmp/envoy.admin http://localhost/healthcheck/fail; sleep 15"]
+                command:
+                - perl
```
skonto, line 94: I like that this is more compact but I feel that we should stick to bash so it is aligned with the rest of the codebase.

config/300-gateway.yaml
norbjd: Thanks for the review. I've tested both; in the `preStop` hook, calling `/drain_listeners?graceful` is similar to calling `/healthcheck/fail`, in the sense both are starting the draining process as you have seen in [Envoy docs](https://www.envoyproxy.io/docs/envoy/v1.26.0/intro/arch_overview/operations/draining#arch-overview-draining). The only difference is that the `/ready` endpoint will now return `503` in the `/healthcheck/fail` case, which is not _that_ important because the pod is already in `Terminating` status when the `preStop` hook is called. 

On **my side**, I feel like `/drain_listeners?graceful` is somewhat clearer than `/healthcheck/fail`, because the former clearly states the goal is to drain, while the latter just makes the healthcheck fail, and people have to read Envoy's doc to know that failing the healthcheck causes starting the drain process.

Anyway, theorically, we can just do the following in the `preStopHook`:

- call `/healthcheck/fail` (or `/drain_listeners?graceful`, similar as I've seen)
- wait (sleep) during `<drain_max_time>` **OR** wait until there is no more in-flight requests

Waiting until all in-flight requests are completed is not simple. We could hack around [like I did](https://github.com/knative-extensions/net-kourier/pull/1203/commits/3e3e933c1100a4c82c751b045a0e5b8678a4a972), but I'm not sure it's the most robust/simple-to-understand solution. So, it might just be easier to wait using `sleep`. The downside of `sleep`ing is that we might wait too much (e.g. all in-flight requests can be completed after 10 seconds, but if we have configured a `<drain_max_time>` of 5 minutes, we will wait 4m50s for nothing). Both solutions have pros and cons.

Another thing we have to be careful is Kubernetes' `terminationGracePeriodSeconds`, that will automatically kill the pod after that period (and it applies also for the `preStopHook`). So, if we have `drain_max_time = 60s`, then `terminationGracePeriodSeconds` must be greater than `60s`, to prevent Kubernetes from killing the pod before the `preStopHook` completes.

That being said, another solution is to use **only** `terminationGracePeriodSeconds`, and in the `preStopHook`:

- call `/healthcheck/fail` (or `/drain_listeners?graceful`, similar)
- sleep for infinity

Then, configure `terminationGracePeriodSeconds` = `<drain_max_time>`. Kubernetes will then be responsible of sending the KILL signal after `drain_max_time`, making the `preStopHook` effectively sleep for `drain_max_time`. Pro of this solution is we just have to configure one thing (`terminationGracePeriodSeconds`), con is we're mixing `terminationGracePeriodSeconds` with the drain process, so it might not be as clear (tell me what you think).

In conclusion from all of this, **my** prefered solution would be:

- define the "drain max time" in an environment variable, so users can easily change it: `DRAIN_TIME_SECONDS`
- define a bigger `terminationGracePeriodSeconds` (compared with `DRAIN_TIME_SECONDS`, e.g. `300`). If people want to have a bigger `DRAIN_TIME_SECONDS`, they must consider increasing this too
- in the `preStopHook`:
  - use `/drain_listeners?graceful`: this is just a personal preference; even if it's similar to `/healthcheck/fail`, as I said in the beginning, I feel like it's clearer to understand, because this command is **made to drain** in-flight requests
  - sleep during `DRAIN_TIME_SECONDS`. In the future, we can just wait until all in-flight requests are done, to avoid waiting for nothing. If you're happy, I can do that in a separate PR to avoid mixing too many things in this one

Tell me what you think; I'll adjust the PR accordingly :smile: sorry about the long answer, but I just wanted to explain all possibilities so we can choose the best one.

config/300-gateway.yaml
norbjd: Yep I understand, I have no strong feeling here, so I'll stick with what you prefer :grin:

To be fair, I've tried the `perl` approach, because some people are allergic to "big" shell scripts :grin: if you prefer the shell approach, I've already made this: https://github.com/knative-extensions/net-kourier/compare/main...norbjd:5be667d697ba34c65946ed42bf4b9533cc27b4c5

The only downside is we need to "open" the `/healtcheck/fail` or `/drain_listeners` endpoint on Envoy's admin interface, because there is no way I know of connecting to the `/tmp/envoy.admin` socket with bash utilities embedded in Envoy's image.

config/300-gateway.yaml
norbjd: (also see this comment on my previous MR: https://github.com/knative-extensions/net-kourier/pull/1200#issuecomment-1923168171 who might justify using `/drain_listeners` instead of `/healthcheck/fail`)

config/300-gateway.yaml
norbjd: Hello, I've reverted to pure shell: https://github.com/knative-extensions/net-kourier/pull/1203/commits/423a6da5f2255f73ff844f06ae5692ff4dc64bdc

Tell me what you think :)

config/300-gateway.yaml
skonto: Thank for the update! I will take a look.

skonto:
Hi @norbjd thanks again for your work and patience.

I like the approach taken from the Contour project: https://github.com/projectcontour/contour/issues/145
as it does not depend on configuring timeouts.
Istio has implemented a similar approach with https://github.com/istio/istio/pull/35059. 
I see some steps to get there to some extent.

a) a documented design like https://github.com/projectcontour/contour/blob/main/design/envoy-shutdown.md
b) a sidecar that handles termination like https://github.com/projectcontour/contour/blob/main/cmd/contour/shutdownmanager.go

I like the go approach compared to bash logic for waiting for active connections because it seems a bit "dirtier" in bash this was proposed also here: https://github.com/projectcontour/contour/pull/2177#issuecomment-524359175.
So definitely should be in another PR but I am still thinking about it.

In the mean time we can make things configurable, see 
 https://github.com/projectcontour/contour/pull/2177#issuecomment-516437707 as also this PR tries to do.

Thus we could do:
1) Keep healthcheck fail endpoint as it signals k8s to also remove it from the service and make the endpoint accessible locally only. 

Btw I agree that the envoy image only contains perl so no nc, socat, curl etc :shrug:. However you can expose the admin
interface locally only and not via a pipe and target 127.0.0.1:9001 directly. Would that work? 

clusters:
  - name: service_stats
    connect_timeout: 0.250s
    type: static
    load_assignment:
      cluster_name: service_stats
      endpoints:
        lb_endpoints:
          endpoint:
            address:
              socket_address:
                address: 127.0.0.1
                port_value: 9001
...

admin:
  access_log:
  - name: envoy.access_loggers.stdout
    typed_config:
      "@type": type.googleapis.com/envoy.extensions.access_loggers.stream.v3.StdoutAccessLog
  address:
    socket_address:
      address: 127.0.0.1
      port_value: 9001

2) pass the drain parameter `drain-time-s` with some time that is configurable as you do in the PR.

3) I am not sure about the drain strategy https://github.com/istio/istio/blob/master/pkg/envoy/proxy.go#L130. Istio sets it to immediate.

--drain-strategy <string>
(optional) Determine behaviour of Envoy during the hot restart drain sequence. During the drain sequence, the drain manager encourages draining through terminating connections on request completion, sending “Connection: CLOSE” on HTTP1, and sending GOAWAY on HTTP2.

gradual: (default) The percentage of requests encouraged to drain increases to 100% as the drain time elapses.
immediate: All requests are encouraged to drain as soon as the drain sequence begins.

I guess we want to notify clients asap (like you have it in the PR).

4) terminationGracePeriodSeconds should be larger than the drain period for sure.

5) Add an e2e test that checks the correctness of what we have.

skonto:
@norbjd hi, pls let me know if you are interested in updating the PR.

skonto:
@norbjd gentle ping!

norbjd:
Woops, really sorry, I've been busy at work these last weeks. I'll check again right now!

norbjd:
Hello @skonto,

Sorry about the delay, and thanks for the detailed answer here https://github.com/knative-extensions/net-kourier/pull/1203#issuecomment-1988464105.

I'm not sure if I have correctly understood the next steps though, sorry :/ Do you want me to 1. continue on this PR or 2. abandon it completely to write a design document (like Istio) defining a Go sidecar-based approach (like Contour) to properly drain the connections? If it's the latter, I'm afraid I won't have much time for it now, but I can try :crossed_fingers:

For more context, the reason I have made this PR - even if it's not the most robust solution I must admit - is that today, draining is not properly done: connections are killed 15s after the gateway is terminated, and it's causing us issues. We are probably going to patch our Kourier version in production with this fix in a near future, so that's why I thought it was better to contribute upstream and make others benefit from it. I see this PR only as a first step to fix a bug, and improve the gateway termination process to avoid abrupt HTTP errors for long-running connections.

Anyway, I took any of your points for the current PR to clean it a little bit:

1. If we make the healthcheck fail before draining, I'm afraid Kubernetes will automatically restart the pod after `livenessProbe.periodSeconds * livenessProbe.failureThreshold` (30s for now), thus completely bypassing the graceful drain process. Maybe I'm wrong, but when the pod is in `Terminating` status (draining is in progress thanks to the `preStop` hook), Kubernetes won't [send new requests to it](https://kubernetes.io/docs/tutorials/services/pods-and-endpoint-termination-flow/), so it should not be necessary to make the healthcheck fail explicitly. Regarding the usage of `/tmp/envoy.admin` UNIX socket, you're right, I've replaced it with `127.0.0.1:9901`, it's easier to manage.

2. `drain-time-s` is configurable through an environment variable (`DRAIN_TIME_SECONDS`), so people can easily overwrite it without requiring to change the args.
3. Yes, the immediate strategy is probably the most suited here, as we want to notify clients as soon as possible.
4. Even if the default `terminationGracePeriodSeconds` set by Kubernetes is 30 seconds, I've added it in the gateway manifest to clarify. I can put a comment next to it to tell it **must** be greater than `DRAIN_TIME_SECONDS`, let me know.
5. I've added an e2e test that basically:

- configures `drain-time-s` to `30` and `terminationGracePeriodSeconds` to `60`
- creates a service and an ingress. It's reusing the `timeout` service, so we can pass an `initialTimeout` parameter to simulate long-running requests
- make two HTTP requests, taking respectively 27 seconds and 33 seconds to complete (slightly less and slightly more than drain time)
- check the first HTTP request succeeds (properly drained), while the second one fails (exceed drain time).

---

Also, sorry about the messy Git history (I hope this will be squashed!) and the broken pipelines, I didn't have access to an IDE, nor a Go compiler, nor a Kubernetes environment at my disposal right now :/ So, I had to do it the old way with a text editor, the Go documentation, and push to check if my changes were OK :grin:

Once I have a proper setup to test on, I'll try again with `kind`, but I'm pretty confident with the e2e test I have added right now. I also have to check why some other e2e tests are failing, but it's probably a weird side effect as they pass on some pipelines.

I'm also available on the CNCF Slack if you want to discuss more informally about this PR :smile:

Thanks!

test/gracefulshutdown/gracefulshutdown_test.go
```diff
@@ -0,0 +1,176 @@
...
+	reqURL := fmt.Sprintf("http://%s.example.com?initialTimeout=%d", name, requestTimeout.Milliseconds())
+	req, err := http.NewRequest("GET", reqURL, nil)
+	if err != nil {
+		return 0, fmt.Errorf("error making GET request: %w", err)
+	}
+
+	resp, err := client.Do(req)
+	if err != nil {
+		var errURL *url.Error
+		// When the gateway cuts the connection (after completing the drain process), an EOF error is returned to the client
```
norbjd, line 166: Note for me: I'm really not sure about this behavior, I need to test in real conditions. I would have expected a 5xx error with a `Connection: close` header.

skonto:
Hi @norbjd thanks for getting back to this one.

> I'm not sure if I have correctly understood the next steps though, sorry :/ Do you want me to 1. continue on this PR or 2. abandon it completely to write a design document (like Istio) defining a Go sidecar-based approach (like Contour) to properly drain the connections? If it's the latter, I'm afraid I won't have much time for it now, but I can try

I would like to continue with 1, this PR. Option 2 was a suggestion for future work.

cc @maschmid @ReToCode for reviews.

ReToCode:
>  I can put a comment next to it to tell it must be greater than DRAIN_TIME_SECONDS, let me know.

Yes please :) 

In general I think that approach is good. Could you please also:

- Update the PR description to match the current status of the code
- Make sure all tests pass successfully

skonto:
@norbjd hi, gentle ping. Pls take a look at @ReToCode's suggestions and let's move this one.

skonto:
@norbjd gentle ping

norbjd:
Thanks for the review, I have fixed the last requested changes :heavy_check_mark:

https://github.com/norbjd/net-kourier/compare/7170cb6bca57dd51a3bb1c75582e48de22573fe9...gateway-prestop-hook-wait-until-all-incoming-requests-are-finished

TL;DR:

- add a comment near `terminationGracePeriodSeconds` to ensure users set a value greater than `DRAIN_TIME_SECONDS`
- run the gracefulshutdown e2e test at the end (otherwise I had conflicts between this test and others, this is what caused flaky tests): **all e2e tests are now passing**
- **PR description has been updated**

ReToCode (approved):
/lgtm
/approve

Thanks for doing this @norbjd

PR #1236, opened 2024-04-02
Increase gateway memory limit

ReToCode:
# Changes
- Increase 3scale-kourier-gateway memory limit

Fixes https://github.com/knative/serving/issues/15052

**Release Note**
```release-note
The default memory limit of 3scale-kourier-gateway was increased to prevent OOMKilled on peak load scenarios
```

### DIFF
--- a/config/300-gateway.yaml
+++ b/config/300-gateway.yaml
@@ -114,7 +114,7 @@ spec:
               memory: 200Mi
             limits:
               cpu: "1"
-              memory: 500Mi
+              memory: 800Mi
       volumes:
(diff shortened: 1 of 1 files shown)


ReToCode:
/assign @skonto 

Tested in https://github.com/knative/serving/pull/15060.

skonto:
Do we know why we need this increase, what is the root cause of the overhead?

ReToCode:
That we added more stuff to encryption, we are definitely creating more envoy config with this (thus, just the tls test were flaky).

PR #1267, opened 2024-06-23
Add RetryPolicy on VirtualHost for upstream transient connection issues

norbjd:
# Changes

- :gift: Add RetryPolicy on VirtualHost for upstream transient connection issues 

/kind enhancement

---

For context, we are operating kourier "at scale":

- thousands of knative services
- hundreds of requests/second
- a lot of activity with new ksvc created, updated, deleted continuously

In the gateway access logs, we sometimes see clients requests ending up in 503. These 503 are always mostly accompanied by `UC` response flag, meaning (from the [docs](https://www.envoyproxy.io/docs/envoy/v1.26.8/configuration/observability/access_log/usage#command-operators)):

> Upstream connection termination in addition to 503 response code.

Connections can be terminated for many reasons, but most of the time, these are linked to transient TCP failures (connect timeout, reset, disconnect, etc.). As of now, the only way to deal with these 503 UC errors is to retry on the caller side, which is not really convenient - and not always possible - for callers.

In order to increase robustness and handle these specific transient cases, Envoy allows setting retry policies at [`VirtualHost` level](https://www.envoyproxy.io/docs/envoy/v1.26.8/api-v3/config/route/v3/route_components.proto#envoy-v3-api-field-config-route-v3-virtualhost-retry-policy) and/or [`Route` level](https://www.envoyproxy.io/docs/envoy/v1.26.8/api-v3/config/route/v3/route_components.proto#envoy-v3-api-field-config-route-v3-routeaction-retry-policy). By default, these retry policies are **not** configured by Kourier, so this is why Envoy throws directly 503s to the caller, sadly.

This PR configures the `RetryPolicy` at the `VirtualHost` level (because every VH has multiple routes, it would be cumbersome to define it on every route). The conditions to retry, as explained in [Envoy docs](https://www.envoyproxy.io/docs/envoy/v1.26.8/configuration/http/http_filters/router_filter#config-http-filters-router-x-envoy-retry-on), covers all transient upstream connection failures:

> - `reset`: Envoy will attempt a retry if the upstream server does not respond at all (disconnect/reset/read timeout.)
> - `connect-failure`: Envoy will attempt a retry if a request is failed because of a connection failure to the upstream server (connect timeout, etc.). [...] NOTE: A connection failure/timeout is a the TCP level, not the request level

_Note 1_: BTW, this is also what istio seems to do by default: https://istio.io/latest/docs/concepts/traffic-management/#retries, https://istio.io/latest/docs/reference/config/networking/virtual-service/#HTTPRetry, but I'm not really familiar with it, so I can just trust the docs and what I find on the web.

_Note 2_: It might be tempting to retry on every upstream errors (e.g. 5xx), but it is probably not a good idea, as "real" 5xx sent by the users applications (in the ksvc) might not be retriable and can cause more harm if retried. Here, we will just focus on TCP connection errors.

---

Regarding the changes made on the PR itself: I'm pretty sure setting the retry policy for **every** `VirtualHost` can't be harmful; but, as I don't know your opinion on this, I've hidden it behind an option (`WithRetryOnTransientUpstreamFailure()`, using option pattern).

For now, the option is always **on** (`pkg/generator/ingress_translator.go`), but if you prefer, I can easily make it configurable through Kourier configmap (e.g. if `retry-on-upstream-transient-failures: true` in the config, call the option; otherwise, bypass it). When adding this option, I have also changed `NewVirtualHostWithExtAuthz`, because I didn't want to add `if`s everywhere, and managing this with options is far more convenient.

So, from there, there are 2 paths I can take:

1. configure the retry by default: in that case, I can completely remove the "options" pattern, and just configure the `RetryPolicy` in `NewVirtualHost` method (and `NewVirtualHostWithExtAuthz`, by extension)
2. make the retry configurable by the user: in that case, I will keep the "options" pattern, but I will need to change `translateIngress` signature (and all methods calling it) to include the `WithRetryOnTransientUpstreamFailure()` option only if the user have opted-in in kourier config

Solution 1 is easier to implement but does not guarantee side-effects (just adding retries in case of TCP connection issues should be fine though...), while solution 2 allows to be more configurable and retry can stay disabled by default.

Tell me what you think. Thanks :pray:

**Release Note**

```release-note
Add retry policy on every virtual host for upstream transient connection issues
```

**Docs**

N/A

### DIFF
--- a/pkg/envoy/api/virtual_host.go
+++ b/pkg/envoy/api/virtual_host.go
@@ -21,39 +21,47 @@ import (
 	extAuthService "github.com/envoyproxy/go-control-plane/envoy/extensions/filters/http/ext_authz/v3"
 	"github.com/envoyproxy/go-control-plane/pkg/wellknown"
 	"google.golang.org/protobuf/types/known/anypb"
+	"google.golang.org/protobuf/types/known/wrapperspb"
 )
 
+type VirtualHostOption func(*route.VirtualHost)
+
 // NewVirtualHost creates a new VirtualHost.
-func NewVirtualHost(name string, domains []string, routes []*route.Route) *route.VirtualHost {
-	return &route.VirtualHost{
+func NewVirtualHost(name string, domains []string, routes []*route.Route, options ...VirtualHostOption) *route.VirtualHost {
+	virtualHost := &route.VirtualHost{
 		Name:    name,
 		Domains: domains,
 		Routes:  routes,
+	}
+
+	for _, opt := range options {
+		opt(virtualHost)
+	}
+
+	return virtualHost
+}
+
+func WithExtAuthz(contextExtensions map[string]string) VirtualHostOption {
+	return func(virtualHost *route.VirtualHost) {
+		filter, _ := anypb.New(&extAuthService.ExtAuthzPerRoute{
+			Override: &extAuthService.ExtAuthzPerRoute_CheckSettings{
+				CheckSettings: &extAuthService.CheckSettings{
+					ContextExtensions: contextExtensions,
+				},
+			},
+		})
+
+		virtualHost.TypedPerFilterConfig = map[string]*anypb.Any{
+			wellknown.HTTPExternalAuthorization: filter,
+		}
 	}
 }
 
-// NewVirtualHostWithExtAuthz creates a new VirtualHost with ExtAuthz settings.
-func NewVirtualHostWithExtAuthz(
-	name string,
-	contextExtensions map[string]string,
-	domains []string,
-	routes []*route.Route) *route.VirtualHost {
-
-	filter, _ := anypb.New(&extAuthService.ExtAuthzPerRoute{
-		Override: &extAuthService.ExtAuthzPerRoute_CheckSettings{
-			CheckSettings: &extAuthService.CheckSettings{
-				ContextExtensions: contextExtensions,
-			},
-		},
-	})
-
-	return &route.VirtualHost{
-		Name:    name,
-		Domains: domains,
-		Routes:  routes,
-		TypedPerFilterConfig: map[string]*anypb.Any{
-			wellknown.HTTPExternalAuthorization: filter,
-		},
+func WithRetryOnTransientUpstreamFailure() VirtualHostOption {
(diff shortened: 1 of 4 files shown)


pkg/generator/ingress_translator.go
```diff
@@ -255,20 +255,22 @@ func (translator *IngressTranslator) translateIngress(ctx context.Context, ingre
 		}
 
 		var virtualHost, virtualTLSHost *route.VirtualHost
+		// TODO(norbjd): do we want to enable this by default?
```
norbjd, line 258: I'm waiting for your opinion on this. Either 1. we enable this by default, and we don't need the option logic; or 2. we make it configurable through Kourier configmap.

pkg/envoy/api/virtual_host.go
```diff
@@ -21,39 +21,47 @@ import (
...
-		Routes:  routes,
-		TypedPerFilterConfig: map[string]*anypb.Any{
+		virtualHost.TypedPerFilterConfig = map[string]*anypb.Any{
 			wellknown.HTTPExternalAuthorization: filter,
-		},
+		}
 	}
+}
 
+func WithRetryOnTransientUpstreamFailure() VirtualHostOption {
```
norbjd, line 60: See PR description to have more context about why I've used the option pattern. TL;DR: if we want to make it configurable, it's easier, and avoids having to write all permutations (no ext-authz/no retry, no ext-authz/retry, ext-authz/no retry, ext-authz/retry).

norbjd:
Note: I don't know how to write an integration test for this... I've tried many things to kill TCP connections, mimic a flaky network, but can't have something working consistently. I guess network issues happen when we expect them the least :sweat_smile: The closest I've found to "cut" connections is to use network policies, but alas it's not supported on `kind` (https://github.com/kubernetes-sigs/kind/issues/842) so it won't help in our case.

norbjd:
Putting back in draft because I need to test thoroughly to ensure it works as expected, sorry for the ping!

norbjd:
Closing, because I cannot guarantee that the retry set in this MR works properly :confused: Sorry again for the premature ping to reviewers :bow:

PR #1273, opened 2024-07-24
Use a different 'static' image that supports s390x and ppc

dprotaso:
Updated image is recommended by Chainguard
Details here: https://www.chainguard.dev/unchained/changes-to-static-git-and-busybox-developer-images-2

## Release Note
```release-note
Support s390x/ppc when building our controllers
```


ReToCode (approved):
/lgtm
/approve

PR #1291, opened 2024-09-23
Fix lint errors

ReToCode:
# Changes
- Fix lint errors which prevent passing CI like [here](https://github.com/knative-extensions/net-kourier/pull/1287)

/kind cleanup

Supported kinds are: api-change, bug, cleanup, deprecation, removal, documentation, enhancement, performance

/assign @skonto

### DIFF
--- a/pkg/config/ext_authz.go
+++ b/pkg/config/ext_authz.go
@@ -114,7 +114,7 @@ func init() {
 
 	ExternalAuthz = &ExternalAuthzConfig{
 		Enabled:    true,
-		Cluster:    extAuthzCluster(host, uint32(port), env.Protocol),
+		Cluster:    extAuthzCluster(host, uint32(port), env.Protocol), //#nosec G115
 		HTTPFilter: externalAuthZFilter(&env),
 	}
 }
--- a/pkg/generator/ingress_translator.go
+++ b/pkg/generator/ingress_translator.go
@@ -191,7 +191,7 @@ func (translator *IngressTranslator) translateIngress(ctx context.Context, ingre
 					// If the service is of type ExternalName, we add a single endpoint.
 					typ = v3.Cluster_LOGICAL_DNS
 					publicLbEndpoints = []*endpoint.LbEndpoint{
-						envoy.NewLBEndpoint(service.Spec.ExternalName, uint32(externalPort)),
+						envoy.NewLBEndpoint(service.Spec.ExternalName, uint32(externalPort)), //#nosec G115
 					}
 				} else {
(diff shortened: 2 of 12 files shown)


pkg/config/ext_authz.go
```diff
@@ -114,7 +114,7 @@ func init() {
 
 	ExternalAuthz = &ExternalAuthzConfig{
 		Enabled:    true,
-		Cluster:    extAuthzCluster(host, uint32(port), env.Protocol),
+		Cluster:    extAuthzCluster(host, uint32(port), env.Protocol), //#nosec G115
```
skonto, line 117: Should we avoid this with:
```
diff --git a/pkg/config/ext_authz.go b/pkg/config/ext_authz.go
index bdcd2ea6..b743768c 100644
--- a/pkg/config/ext_authz.go
+++ b/pkg/config/ext_authz.go
@@ -39,7 +39,6 @@ import (
 const (
        extAuthzClusterName = "extAuthz"
        // See https://en.wikipedia.org/wiki/Registered_port.
-       unixMaxPort = 65535
 )
 
 // ExternalAuthz is the configuration of external authorization.
@@ -102,16 +101,11 @@ func init() {
                panic(err)
        }
 
-       port, err := strconv.Atoi(portStr)
+       port, err := strconv.ParseUint(portStr, 10, 16)
        if err != nil {
                panic(err)
        }
 
-       if port > unixMaxPort {
-               // Bail out if we exceed the maximum port number.
-               panic(fmt.Sprintf("port %d bigger than %d", port, unixMaxPort))
-       }
-
        ExternalAuthz = &ExternalAuthzConfig{
                Enabled:    true,
                Cluster:    extAuthzCluster(host, uint32(port), env.Protocol),
```
:thinking:

pkg/generator/ingress_translator.go
```diff
@@ -191,7 +191,7 @@ func (translator *IngressTranslator) translateIngress(ctx context.Context, ingre
 					// If the service is of type ExternalName, we add a single endpoint.
 					typ = v3.Cluster_LOGICAL_DNS
 					publicLbEndpoints = []*endpoint.LbEndpoint{
-						envoy.NewLBEndpoint(service.Spec.ExternalName, uint32(externalPort)),
+						envoy.NewLBEndpoint(service.Spec.ExternalName, uint32(externalPort)), //#nosec G115
```
skonto, line 194: We can't avoid that https://discuss.kubernetes.io/t/why-are-int32-s-all-over-the-place-where-uint32-s-are-apparently-more-appropriate-number-of-port-etc/21882 :D

PR #1311, opened 2024-12-04
Allow to disable envoy server header injection

skonto:
# Changes

- Downstream users want to hide the [server header](https://security.stackexchange.com/questions/23256/what-is-the-http-server-response-header-field-used-for). If there is no such header present in the response, by default Envoy injects its own that has the value `envoy` (based on the default server_name config). Envoy allows to configure when the server header is injected. Here we support only the option to prevent Envoy's header injection. In a secured environment probably this should be on by default.

/kind <kind>

Fixes #

**Release Note**

```release-note
A new configuration option `disable-envoy-server-header` prevents Envoy server header from being injected when there is none such header in the response.
```


ReToCode (approved):
/lgtm
/approve

PR #1326, opened 2025-02-13
fix: update-codegen.sh

ystkfujii:
# Changes

- I fixed an error that occurred when running `./hack/update-codegen`. 
  - This is the same fix as https://github.com/knative/eventing/pull/8183

<details>

```
$ ./hack/update-codegen.sh
...
...
--- Deepcopy Gen
unknown shorthand flag: 'O' in -O
Usage of /tmp/knative.8bJLBWxQ/go/bin/deepcopy-gen:
      --add_dir_header                   If true, adds the file directory to the header of the log messages
      --alsologtostderr                  log to standard error as well as files (no effect when -logtostderr=true)
      --bounding-dirs strings            Comma-separated list of import paths which bound the types for which deep-copies will be generated.
      --go-header-file string            the path to a file containing boilerplate header text; the string "YEAR" will be replaced with the current 4-digit year
      --log_backtrace_at traceLocation   when logging hits line file:N, emit a stack trace (default :0)
      --log_dir string                   If non-empty, write log files in this directory (no effect when -logtostderr=true)
      --log_file string                  If non-empty, use this log file (no effect when -logtostderr=true)
      --log_file_max_size uint           Defines the maximum size a log file can grow to (no effect when -logtostderr=true). Unit is megabytes. If the value is 0, the maximum file size is unlimited. (default 1800)
      --logtostderr                      log to standard error instead of files (default true)
      --one_output                       If true, only write logs to their native severity level (vs also writing to each lower severity level; no effect when -logtostderr=true)
      --output-file string               the name of the file to be generated (default "generated.deepcopy.go")
      --skip_headers                     If true, avoid header prefixes in the log messages
      --skip_log_headers                 If true, avoid headers when opening log files (no effect when -logtostderr=true)
      --stderrthreshold severity         logs at or above this threshold go to stderr when writing to files and stderr (no effect when -logtostderr=true or -alsologtostderr=true) (default 2)
  -v, --v Level                          number for the log level verbosity
      --vmodule moduleSpec               comma-separated list of pattern=N settings for file-filtered logging
unknown shorthand flag: 'O' in -O
...
...
```

</details>

/kind bug


pkg/reconciler/ingress/config/store.go
```diff
@@ -28,7 +28,7 @@ import (
 type cfgKey struct{}
 
 // Config contains the configmaps requires for revision reconciliation.
-// +k8s:deepcopy-gen=false
+// +k8s:deepcopy-gen=true
```
ystkfujii, line 31: It looks to me that DeepCopy is being used.

https://github.com/knative-extensions/net-kourier/blob/dcf76e1d4f84d9f8427c86344b8c65ba173ef5e1/pkg/generator/ingress_translator_test.go#L971

ystkfujii:
@dprotaso @skonto Could you review the PR ?

dprotaso:
/lgtm
/approve

cc @skonto

pkg/reconciler/ingress/config/store.go
dsimansk: 🤔 it might have been just generated once...

PR #1327, opened 2025-02-13
fix(config): Update the controller DNS in bootstrap

arsenetar:
# Changes

Update the controller DNS name to use the short form of the DNS that should be resolvable within the namespace as all components involved should be in the same namespace.

/kind bug

This fixes issues if deployed outside of `knative-serving` namespace (via the operator etc.) as the original is hardcoded with that namespace.  As in that case the gateways will never become healthy as they cannot communicate with the incorrect address in this configuration.  The short form change here should be functional in all cases.

**Release Note**

```release-note
Fix hardcoded namespace in DNS for kourier controller in envoy bootstrap
```

### DIFF
--- a/config/200-bootstrap.yaml
+++ b/config/200-bootstrap.yaml
@@ -113,7 +113,7 @@ data:
                 endpoint:
                   address:
                     socket_address:
-                      address: "net-kourier-controller.knative-serving"
+                      address: "net-kourier-controller"
                       port_value: 18000
           type: STRICT_DNS
     admin:


arsenetar:
Hmm, the manifests here do a dual namespace install vs the operator, which would need it, figured this would be the right place to update perhaps patching the manifest in just the operator's version makes more sense.

skonto:
Hi @arsenetar I think this should be managed at the operator side. cc @houshengbo 
I am curious if it would even work without a namespace.

arsenetar:
> Hi @arsenetar I think this should be managed at the operator side. cc @houshengbo I am curious if it would even work without a namespace.

It does for the deployments by the operator where its all inside the same ns in the clusters we encountered the issue.

I'll be moving this over to the operator side.

Issue #1346, opened 2025-05-09
net-kourier-controller goes into CrashLoopBackOff due to AKS API server throttling


dg-mrenaud:
### tl;dr
Every once in a while, the `net-kourier-controller` pod goes into a CrashLoopBackOff because it tries to retrieve all endpoints at once, gets throttled by the AKS API server and is unable to recover until almost all `ksvc` and `king` resources are deleted.

### Versions
- AKS: `1.31.6` (Standard pricing tier)
- Knative Serving: `1.17.0`
- Kourier: `sha256:ab7885b64daf628a6b1e8f0aef572ec6752cd0497172963ffeec80cd870f5fcc`

### Description
We have a Knative Serving setup on our test AKS cluster that we use to create frontend previews generated by pull request validation builds - these get cleaned up regularly to keep the number of `ksvc` resources in check, though a couple hundred are usually present at a time, mostly around 300 or so.

At seemingly random intervals, however, the `net-kourier-controller` pod crashes, rendering the services inaccessible. When this happens, the pod log usually looks something like this:
```
2025/05/09 08:56:04 Registering 2 clients
2025/05/09 08:56:04 Registering 4 informer factories
2025/05/09 08:56:04 Registering 5 informers
2025/05/09 08:56:04 Registering 1 controllers
{"severity":"INFO","timestamp":"2025-05-09T08:56:04.891210919Z","logger":"net-kourier-controller","caller":"profiling/server.go:65","message":"Profiling enabled: false","commit":"bfecb96-dirty"}
{"severity":"INFO","timestamp":"2025-05-09T08:56:06.868086592Z","logger":"net-kourier-controller","caller":"leaderelection/context.go:47","message":"Running with Standard leader election","commit":"bfecb96-dirty"}
{"severity":"INFO","timestamp":"2025-05-09T08:56:10.131139801Z","logger":"net-kourier-controller","caller":"ingress/controller.go:222","message":"Priming the config with 269 ingresses","commit":"bfecb96-dirty"}
[...]

PR #1349, opened 2025-05-28
WIP: cache ingress resources during startup

dsimansk:
# Changes

- WIP: cache ingress resources during startup

In the worst case scenario we could disable prewarming of Ingresses with config flag to be processed in reconciliation loop during runtime. It seems a bit unfortunate design decision that warmup might prevent controller to start.

Related to #1346


dsimansk:
I'd rather use informers in the startup sequence. Even if there's any missing ingress at the startup, it's "prewarm" and eventually ingresses are reconciled over time. At the current impl there are at least 4 api queries per existing Ingress in the cluster.

PR #1350, opened 2025-05-30
Align GH actions to Serving common patterns

dsimansk:
# Changes

- Rm slack notification to old slack instance
- Add log collection on failure in e2e tests
- Adjust e2e triggers to common pattern

/cc @Fedosin @dprotaso

### DIFF
--- a/.github/workflows/kind-e2e-upgrade.yaml
+++ b/.github/workflows/kind-e2e-upgrade.yaml
@@ -69,5 +69,5 @@ jobs:
       if: ${{ failure() }}
       with:
         cluster-resources: nodes,namespaces,crds
-        namespace-resources: configmaps,pods,svc,ksvc,route,configuration,revision,king
+        namespace-resources: configmaps,pods,svc,king
         artifact-name: logs-${{ matrix.k8s-version}}-${{ matrix.upstream-traffic }}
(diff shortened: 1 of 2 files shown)


.github/workflows/kind-e2e-upgrade.yaml
```diff
@@ -66,3 +63,11 @@ jobs:
...
       run: |
         ./test/e2e-upgrade-kind.sh
+
+    # Collect debug logs on failure
+    - uses: chainguard-dev/actions/kind-diag@main
+      if: ${{ failure() }}
+      with:
+        cluster-resources: nodes,namespaces,crds
+        namespace-resources: configmaps,pods,svc,king
+        artifact-name: logs-${{ matrix.k8s-version}}-${{ matrix.upstream-traffic }}
```
ystkfujii, line 73: While I've never actually used it myself, I'm LGTM based on the following points:
- It's also used in other repositories.
  - https://github.com/knative-extensions/net-contour/blob/main/.github/workflows/kind-e2e.yaml#L47
- From a quick glance, it seems to only handle read operations.
  - https://github.com/chainguard-dev/actions/blob/main/kind-diag/action.yaml
- Even if this step crashes due to errors, it shouldn't cause any major issues.

ystkfujii:
LGTM !!!
However, please note that I don't have the necessary permissions for this.

dsimansk:
PTAL, cleanup of few parts per description.

/assign @dprotaso

PR #1351, opened 2025-05-31
feat: Enabled customization of the access log format

ystkfujii:
# Changes

-
-
-

/kind <kind>

Fixes #

**Release Note**

```release-note

```

**Docs**

```docs

```


dsimansk (changes requested):
Thanks a lot, look great! Only ConfigMap example is missing.

ystkfujii:
@dsimansk 

Ah... sorry, I forgot again 🙇 
I’ve fixed it now. Please take a look when you have a moment.

Also, could you please rerun the failed GitHub Actions test? I think it will succeed if you try again. 
Unfortunately, it seems that I don't have permission to trigger a re-run myself 😢

dsimansk (approved):

PR #1352, opened 2025-05-31
refactor: Support for loading certificate settings from configmap

ystkfujii:
# Changes

- :gift: Support for loading certificate settings from configmap.

/kind cleanup

Ref: https://github.com/knative-extensions/net-kourier/issues/877#issuecomment-1425846828

**Release Note**

```release-note
Support for loading certificate settings from configmap.
```

### DIFF
--- a/config/200-config.yaml
+++ b/config/200-config.yaml
@@ -109,3 +109,7 @@ data:
     # Defaults to false. More info Envoy Docs.
     # see: https://www.envoyproxy.io/docs/envoy/latest/api-v3/extensions/filters/http/ext_authz/v3/ext_authz.proto.html#extensions-filters-http-ext-authz-v3-buffersettings
     extauthz-pack-as-byte: "false"
+
+    # Specifies the secret that contains the TLS certificate and key pair when using HTTPS communication with Kourier Ingress.
+    certs-secret-name: ""
+    certs-secret-namespace: ""
--- a/config/200-config.yaml
+++ b/config/200-config.yaml
@@ -85,31 +85,39 @@ data:
     disable-envoy-server-header: "false"
 
     # The external authorization service and port, my-auth:2222.
+    # This overrides the corresponding environment variable if it is set.
     extauthz-host: ""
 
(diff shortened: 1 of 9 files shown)


ystkfujii:
Please review when you have time.

config/200-config.yaml
```diff
@@ -109,3 +109,7 @@ data:
     # Defaults to false. More info Envoy Docs.
     # see: https://www.envoyproxy.io/docs/envoy/latest/api-v3/extensions/filters/http/ext_authz/v3/ext_authz.proto.html#extensions-filters-http-ext-authz-v3-buffersettings
     extauthz-pack-as-byte: "false"
+
+    # Specifies the secret that contains the TLS certificate and key pair when using HTTPS communication with Kourier Ingress.
+    certs-secret-name: ""
+    certs-secret-namespace: ""
```
dsimansk, line 115: I'd like to mention that env vars are still supported, but `ConfigMap` will take priority and override env var values.

Vice versa I wonder if we should add comment to deployment wrt/ possibility to use CM now. More like brainstorming open ended questions. Wdyt?

ystkfujii: I agree. Without reading the code, the current behaviour isn't clear, so I’ve added comments to note that ConfigMap values override environment variables.

In theory, mentioning this in the deployment makes sense, but in practice, comments there get lost during 'kubectl apply', and I suspect few users notice them.

Also, is using environment variables considered deprecated in Kourier?
If so, we could make that clearer, perhaps with a log message when environment variables are used, or by noting it in the README.

test/e2e-kind.sh
```diff
@@ -74,8 +74,33 @@ go test -v -tags=e2e ./test/tls/... \
 
 kubectl -n "${KOURIER_CONTROL_NAMESPACE}" patch configmap/config-kourier --type merge -p '{"data":{"cipher-suites":""}}'
 
-echo ">> Setup one wildcard certificate"
+echo ">> Setup one wildcard certificate from configmap"
```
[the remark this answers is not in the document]
ystkfujii: Fixed

dsimansk (approved):

PR #1353, opened 2025-06-02
Refactor e2e-kind.sh to use rollout status instead of condition check

ystkfujii:
# Changes

:bug: Refactoring performed on e2e-kind.sh:
- 1: Removed unnecessary namespace specification.
  - This should have been fixed in https://github.com/knative-extensions/net-kourier/pull/1334 🙇 
- 2: Modified to reference rollout status instead of condition.

Regarding point 2:

We are experiencing intermittent failures with GitHub Actions.
- https://github.com/knative-extensions/net-kourier/actions/runs/15393564244/job/43308474380
- https://github.com/knative-extensions/net-kourier/actions/runs/15344106223/job/43176455027

From what I’ve observed, the following behaviour occurs:
- GitHub Actions fail during tests that configure path prefixes.
- Until the path prefix configuration is reflected in the Pod (externalAuthz), a 403 error occurs.

The `condition=Available` check likely doesn't guarantee that the pod has completed its rollout.
Therefore, I have modified the configuration to verify the rollout status instead.

/kind bug

### DIFF
--- a/test/e2e-kind.sh
+++ b/test/e2e-kind.sh
@@ -110,7 +110,7 @@ kubectl -n "${KOURIER_CONTROL_NAMESPACE}" patch configmap/config-kourier --type 
     "extauthz-protocol": "grpc"
   }
 }'
-kubectl -n "${KOURIER_CONTROL_NAMESPACE}" rollout restart -n knative-serving deployment/net-kourier-controller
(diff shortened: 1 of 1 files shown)


dsimansk:
@ystkfujii thanks!

PR #1360, opened 2025-07-19
Refactor: Use config embedded in context

ystkfujii:
# Changes

- Removed the extAuthz field in Reconciler, instead utilizing the Config.Kourier.ExternalAuthz embedded within the context.

/kind cleanup


linkvt:
Not an approver in this repo but I'll leave an /lgtm as the change looks good 👍 
/lgtm

PR #1375, opened 2025-09-09
Added support to configure listener ip addresses with comma separated list of IPs (ex: "0.0.0.0,::")

### LINKED ISSUE #1299: Add support for IPV6 listeners
When deploying kourier, pod does not get into ready state (even after overwriting bootstrap to ipv6 addresses) because dynamic listeners added by net-kourier explicitly and only listen on ipv4.

https://github.com/knative-extensions/net-kourier/blob/c7434e420928104fcf51bf6c6fe50b0a20aed426/pkg/envoy/api/listener.go#L178C16-L178C23

agallegoz:
# Changes

- Added new configmap param "listen-ip-addresses" to configure the list of IP address to listen to (ex: "0.0.0.0,::") for listening IPv4 and IPv6. Default behaviour is "0.0.0.0" to keep backward compatibility
- Added tests for new functionality
- Modified previous tests to include the new arguments

/kind <kind>enhancement

Fixes #1299

**Release Note**

```release-note
Add ability to set a list of listening IP addresses. This enables the usage of kourier in clusters with IPv4 and IPv6.
```

**Docs**

```docs

```

### DIFF
--- a/pkg/reconciler/ingress/config/kourier.go
+++ b/pkg/reconciler/ingress/config/kourier.go
@@ -24,6 +24,7 @@ import (
 	"os"
 	"strconv"
 	"time"
+	"strings"
 
 	"github.com/kelseyhightower/envconfig"
 	corev1 "k8s.io/api/core/v1"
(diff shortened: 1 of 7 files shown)


agallegoz:
@dsimansk @skonto another gently ping

pkg/reconciler/ingress/config/kourier.go
```diff
@@ -56,6 +57,9 @@ const (
 	// enableCryptoMB is the config map for enabling CryptoMB private key provider.
 	enableCryptoMB = "enable-cryptomb"
 
+	// listenIPAddress receives a the list of IP addresses to listen to.
```
linkvt:
```suggestion
	// listenIPAddresses receives the list of IP addresses to listen to.
```

PR #1394, opened 2025-11-17
[wip] Add W3C trace context propagation support

### LINKED ISSUE #1393: Add W3C Trace Context support for distributed tracing
Kourier currently only supports B3 headers for distributed tracing propagation. 
This limits interoperability with knative itself which does not support extracting context from the old zipkin headers anymore but only from the W3C Trace Context standard (traceparent/tracestate headers).

This behavior is already in place since knative 1.19, see:
- https://github.com/knative/pkg/blob/7e6a0208dd367d18db5fe2956b3ad151f252f941/observability/tracing/provider.go#L47-L52
- https://github.com/knative/serving/pull/15961

## Proposal:

Add a new configuration option to the `config-kourier` ConfigMap with two values:
- `b3` - B3 headers only (backward compatible)
- `b3-w3c` - Dual propagation: extracts from B3 first (fallback to W3C), injects both header formats (default)

This is possible thanks to the new `TraceContextOption` field added in envoy 1.36: 
- https://www.envoyproxy.io/docs/envoy/v1.36.2/api-v3/config/trace/v3/zipkin.proto.html#envoy-v3-api-field-config-trace-v3-zipkinconfig-trace-context-option
- https://www.envoyproxy.io/docs/envoy/latest/version_history/v1.36/v1.36.0.html

That would mean we:
- enable interoperability with knative itself and other W3C-compliant observability tools
- maintain backward compatibility
- default to dual propagation for better compatibility

/kind enhancement
/assign

linkvt:
# WIP
- will check again if we can use some kind of native implementation instead of the b3+w3c workaround

# Changes

- Add configurable W3C trace context propagation trough the new `tracing-context-option` setting in the `config-kourier` ConfigMap
  - Default to the current behavior which is backward compatible. I would like to default to using B3+W3C but this does only work with envoy 1.36...
  - using B3+W3C allows tracing again since knative 1.19 only supports W3C trace context and not B3 anymore
  - see #1393 for more details
- Upgrade envoy to 1.36 (requirement for the above change)

/kind enhancement

Fixes #1393 

## Release Note

```release-note
Add configurable W3C trace context propagation trough the new `tracing-context-option` setting in the `config-kourier` ConfigMap:
  - "b3" - B3 headers only (backward compatible default)
  - "b3-w3c" - B3 + W3C traceparent headers
```

### DIFF
--- a/config/200-config.yaml
+++ b/config/200-config.yaml
@@ -126,3 +126,8 @@ data:
     # This value overrides environment variable if defined.
     certs-secret-name: ""
     certs-secret-namespace: ""
+
+    # Trace context propagation format for distributed tracing (default: "b3"):
+    # - "b3": B3 headers only (x-b3-traceid, x-b3-spanid, etc.)
+    # - "b3-w3c": B3 + W3C traceparent headers (extracts B3 first, injects both) - requires Envoy 1.36+
+    tracing-context-option: "b3"
--- a/pkg/envoy/api/http_connection_manager.go
+++ b/pkg/envoy/api/http_connection_manager.go
@@ -117,11 +117,17 @@ func NewHTTPConnectionManager(routeConfigName string, kourierConfig *config.Kour
 	if kourierConfig.Tracing.Enabled {
 		mgr.GenerateRequestId = wrapperspb.Bool(true)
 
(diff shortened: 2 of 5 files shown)


linkvt:
Superseded by #1395

PR #1395, opened 2025-11-21
Migrate to OTLP and remove Zipkin tracing

### LINKED ISSUE #1393: Add W3C Trace Context support for distributed tracing

linkvt:
# Changes

- Migrate from Zipkin to OTLP (OpenTelemetry Protocol) for distributed tracing
  - Add support for both HTTP and gRPC OTLP protocols
  - Improve the tracing test coverage quite a lot
- Update Envoy to version 1.36
- Remove all Zipkin-specific dependencies and code
- Remove Maistra proxy and Envoy v2 ext_authz support
  - Maistra 2.4 is no longer maintained and the project has been discontinued. Envoy v2 API was removed in v1.18.0 (April 2021). All tested Envoy versions (1.34+) only support v3 API.

/kind enhancement

Fixes #1393  
Supersedes #1394 

## Release Note

```release-note
Migrated distributed tracing from Zipkin to OpenTelemetry Protocol (OTLP).

Action required: Users with tracing enabled must update their `config-kourier` ConfigMap to use the new OTLP configuration. The old `tracing-collector-full-endpoint` key has been removed. Use `tracing-endpoint`, `tracing-protocol`, and `tracing-sampling-rate` instead.
```

**Docs**

```docs

```
I didn't find anything remotely related in the docs so I'm not sure how we want to handle that.

/cc @dprotaso might make sense that you give it a look after implementing Otel in serving etc


.github/workflows/kind-e2e.yaml
```diff
@@ -21,10 +21,9 @@ jobs:
         - v1.33.x
 
         gateway:
-        - quay.io/maistra-dev/proxyv2-ubi8:2.4-latest
```
dsimansk: We kept maistra image here for a compatibility check with Serverless versions.

linkvt: I thought so and did some research (see PR description), the project didn't really seem alive anymore as also indicated by this blog post thats more than 2 years old:
> Thus, to better serve our customers and the Istio community, the time has come to move away from our Maistra.io distribution of Istio and to deliver an OpenShift Service Mesh that is based directly on the Istio project.
> https://www.redhat.com/en/blog/introducing-a-new-operator-for-istio-on-openshift

Do we still want to include it in our CI tests?

PR #1456, opened 2026-05-19
use controller's context in xds_server

maschmid:
# Changes
- :broom: Pass controller's context to XdsServer

so that the `grpcServer.GracefulStop()` is actually invoked (as otherwise there is nothing that would cancel the currently used `context.Background()` context)

### DIFF
--- a/pkg/envoy/server/xds_server.go
+++ b/pkg/envoy/server/xds_server.go
@@ -44,8 +44,7 @@ type XdsServer struct {
 	snapshotCache  cache.SnapshotCache
 }
 
-func NewXdsServer(managementPort uint, callbacks xds.Callbacks) *XdsServer {
-	ctx := context.Background()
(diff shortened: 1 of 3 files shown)


pkg/envoy/server/xds_server.go
```diff
@@ -44,8 +44,7 @@ type XdsServer struct {
 	snapshotCache  cache.SnapshotCache
 }
 
-func NewXdsServer(managementPort uint, callbacks xds.Callbacks) *XdsServer {
-	ctx := context.Background()
+func NewXdsServer(ctx context.Context, managementPort uint, callbacks xds.Callbacks) *XdsServer {
 	snapshotCache := cache.NewSnapshotCache(true, cache.IDHash{}, nil)
 	srv := xds.NewServer(ctx, snapshotCache, callbacks)
```
linkvt, line 49: I think we should launch the server with context.Background to not interrupt pending requests when the parent context is cancelled while using the parent context to trigger the graceful shutdown in RunManagementServer

